PR #8, opened 2017-08-24
Validate `manifest.item` during `RollCall` phase.

### LINKED ISSUE #7: Make sure manifest.items are in the database as events votes
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L761

type is 'RollCall', we already have the events votes from the peer
in the database, so we just have to make sure that those listed in the
manifest also exist in our database or else the manifest is bogus
TODO: make sure manifest.items are in the database as events votes

mattcollier:
Fixes #7.


lib/voteStorage.js
```diff
@@ -113,6 +125,24 @@ api.add = (ledgerNodeId, electionTopic, vote, options, callback) => {
...
+ */
+api.exists = (ledgerNodeId, voteHash, callback) => {
+  const query = {
+    ledgerNodeId: database.hash(ledgerNodeId),
+    'meta.deleted': {$exists: false},
+    'meta.voteHash': voteHash
+  };
+  const collection = database.collections.continuity2017_vote;
+  // FIXME: is there a better/faster way?
+  collection.findOne(query, {_id: 1}, (err, result) => callback(err, !!result));
```
mattcollier: I was specifically avoiding using `count` due to the issues we discussed when implementing it in another module in the past.  I believe it is expensive and unreliable in some cases?

dlongley: If it can use an index, it should be fast. But aggregate is also an option (which would also use an index). I would think either of these would be better than `findOne` x 100, etc.

lib/election.js
```diff
@@ -767,22 +767,33 @@ function _getManifestFromPeer(
...
-        const manifest = results.manifest;
         return _validateEvents(
           ledgerNode, manifest.item, manifest.blockHeight, callback);
       }
-      callback();
+      // type === 'RollCall', we already have the events votes from the peer
+      // in the database, so we just have to make sure that those listed in the
+      // manifest also exist in our database or else the manifest is bogus
+      async.every(manifest.item, (voteHash, callback) =>
+        api._storage.votes.exists(ledgerNode.id, voteHash, callback),
```
mattcollier: idk, is it possible to have a the same `voteHash` for documents that are have different `electionTopic`s?  I would think at the signature would be different on those documents at least, making the hash unique?

dlongley: I guess that vote hashes can only get stored if what the hash was on was verified -- meaning it would have to be different data for a vote on voters (rollcall) vs. a vote on manifests. So it's probably ok to omit it. That's fine, let's omit, no change here.

mattcollier:

PR #24, opened 2017-08-25
Check for existing vote before certification.

### LINKED ISSUE #16: If the vote manifest is already stored, skip certification
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L212

TODO: should be able to see if the vote manifest is already stored
and skip certification if so

@dlongley does this mean...

That somewhere before [L215](https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L215) 

1. Hash the `vote`
2. Use the new `exists` API (in PR) to check for existence of the vote.
3. If it does exist, return early (inside the `async.each`), skipping the `_verifyVote` and `api._storage` steps?

mattcollier:
Fixes #16.

Prior to this optimization, the mutiNode Consensus test (writing 4 blocks) took a total of 24s.  After optimization 16s.  So that's a significant 33% improvement!

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -209,35 +209,39 @@ api.isBlockElector = (voter, electors) => {
  * @param callback(err) called once the operation completes.
  */
 api.certify = (ledgerNode, electionTopic, votes, callback) => {
-  // TODO: should be able to see if the vote manifest is already stored
-  //   and skip certification if so
   async.each(votes, (vote, callback) => {
-    _verifyVote(ledgerNode, vote, electionTopic, err => {
-      if(err) {
-        // ignore failed votes, do not store them
-        logger.verbose('Non-critical error in _verifyVote.', err);
-        return callback();
-      }
-      // TODO: reject votes with more than X number of recommended electors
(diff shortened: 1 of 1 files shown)


lib/election.js
```diff
@@ -209,35 +209,38 @@ api.isBlockElector = (voter, electors) => {
...
-      api._hasher(vote, (err, hash) => {
-        if(err) {
-          return callback(err);
+    async.auto({
+      voteHash: callback => api._hasher(vote, callback),
+      voteExists: ['voteHash', (results, callback) =>
+        api._storage.votes.exists(ledgerNode.id, results.voteHash, callback)],
+      verify: ['voteExists', (results, callback) => {
+        if(results.voteExists === true) {
+          return callback();
```
dlongley: Please add a comment above `return callback`:

```js
// vote already exists, do not need to verify or store again
```

mattcollier: @dlongley done, ready for another pass.

PR #25, opened 2017-08-25
Fix18

mattcollier:
@dlongley this is not finished, but can you look it over now to see if it's doing the right thing?

If this patch is OK, some other new issues are exposed with this patch.  Possible race conditions exposed.

See the [comment](https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/compare/fix18#diff-c945a46d13b34fcaff544d966cffcabaR308)


mattcollier:
@dlongley you can see the aforementioned errors on the CI build.  Although the CI build passed, local tests are failing intermittently.  I want to verify the validity of the patch before investigating that issue further.

PR #26, opened 2017-08-25
Add latest voters to list of recommended electors.

### LINKED ISSUE #23: Include all current electors
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L423

FIXME: include all current electors (should filter this by those
that voted in the last election so we don't recommend those that
aren't participating)

mattcollier:
Fixes #23.

Note, that the `_recommendElectors` function is used in the basic API tests, but it is overridden in the multiNode test.

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -21,6 +21,9 @@ require('./config');
 
 jsigs.use('jsonld', brDidClient.jsonld);
 
+// maximum number of electors if not specified in the ledger configuration
+const MAX_ELECTOR_COUNT = 10;
+
 // module API
 const api = {};
 module.exports = api;
@@ -48,20 +51,7 @@ api._recommendElectors = _recommendElectors;
  */
 api.getBlockElectors = (ledgerNode, blockHeight, callback) => {
   async.auto({
-    config: callback =>
-      ledgerNode.storage.events.getLatestConfig((err, result) => {
-      if(err) {
-        return callback(err);
-      }
-      const config = result.event.input[0];
-      if(config.consensusMethod !== 'Continuity2017') {
-        return callback(new BedrockError(
-          'Consensus method must be "Continuity2017".', 'InvalidStateError', {
-            consensusMethod: config.consensusMethod
-          }));
-      }
-      callback(null, config);
-    }),
+    config: callback => _getLatestConfig(ledgerNode, callback),
     latestBlock: callback =>
       ledgerNode.storage.blocks.getLatest((err, result) => {
         if(err) {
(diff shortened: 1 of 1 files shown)


lib/election.js
```diff
@@ -873,6 +880,26 @@ function _createVote(ledgerNode, electionTopic, vote, voter, callback) {
   }, (err, results) => err ? callback(err) : callback(null, results.store));
 }
 
+function _getLatestConfig(ledgerNode, callback) {
+  ledgerNode.storage.events.getLatestConfig((err, result) => {
+    if(err) {
+      return callback(err);
+    }
+    // `getLatesConfig` returns an empty object before genesis block is written
```
dlongley: typo `getLatesConfig`

lib/election.js
```diff
@@ -417,32 +408,48 @@ function _createEventManifest(ledgerNode, blockHeight, callback) {
  *          completes.
  */
 function _recommendElectors(ledgerNode, voter, electors, manifest, callback) {
-  // TODO: use given parameters and blockchain to determine a set of
-  // electors to recommend for the next block
```
dlongley: Let's keep this TODO around for later.

lib/election.js
```diff
@@ -417,32 +408,48 @@ function _createEventManifest(ledgerNode, blockHeight, callback) {
...
 
   // TODO: implement option to add all nodes that sent in events to the
   // previous block to the elector pool that can be contacted
 
   async.auto({
+    config: callback => _getLatestConfig(ledgerNode, callback),
     latestBlock: callback => ledgerNode.storage.blocks.getLatest(callback),
-    getEventPeers: ['latestBlock', (results, callback) => {
+    latestVoters: ['latestBlock', (results, callback) => {
+      // there are no election results in the genesis block
```
dlongley: This isn't true? There are election results in the genesis block? At least that's what my database says.

lib/election.js
```diff
@@ -417,32 +408,48 @@ function _createEventManifest(ledgerNode, blockHeight, callback) {
...
+          .map(v => v.voter);
+      recommendedElectors.push(...voters);
+      callback();
+    }],
+    eventPeers: ['latestVoters', (results, callback) => {
       if(!results.latestBlock.eventBlock.block) {
         return callback(null, []);
       }
       const events = results.latestBlock.eventBlock.block.event || [];
       // TODO: _getEventPeers returns an array...
```
dlongley: What's TODO here? This is just a statement of truth.

lib/election.js
mattcollier: Maybe what I was seeing was *before* genesis block was written then.

lib/election.js
mattcollier: I was wondering the same thing, I'll remove it.

lib/election.js
mattcollier: @dlongley `ledgerNode.storage.blocks.getLatest`  returns: 
```
eventBlock: {}
```
While working on block 1, since there are no `event` blocks yet.  

Changing the comments.

lib/election.js
mattcollier: I'll put it back in, but what does that comment mean?  Seems like a description of what the function is doing already to me.

mattcollier:
@dlongley Ready for another pass.

lib/election.js
dlongley: Yeah, it does describe what the function is already doing, so we should be more specific:

```js
// TODO: determine if a better deterministic algorithm (better == more secure, performant)
// could be used here that uses given parameters and blockchain data to recommend
// electors for the next block
```

mattcollier: done.

PR #27, opened 2017-08-25
Ensure that peer manifest contains required majority.

### LINKED ISSUE #12: Ensure that manifest.items has >= twoThirds of electors.length
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L774

TODO: must also ensure that manifest.items has >= twoThirds of
electors.length... otherwise the roll-call is insufficient to get
consensus!

@dlongley does this mean we do the same type of calculation done here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L113

Which involves getting the `latestBlock` and counting the votes that were cast on the last block?  As here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L113

mattcollier:
Fixes #12.

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -771,7 +771,26 @@ function _getManifestFromPeer(
     },
     manifest: callback => api._client.getManifest(
       manifestHash, peerId, callback),
-    sync: ['hashes', 'manifest', (results, callback) => {
+    checkMajority: ['manifest', (results, callback) => {
+      const manifest = results.manifest;
+      const blockHeight = manifest.blockHeight;
+      const manifestItemCount = manifest.item.length;
+
+      // TODO: consider passing electors through to optimize away this look up
+      api.getBlockElectors(ledgerNode, blockHeight, (err, result) => {
+        if(err) {
+          return callback(err);
+        }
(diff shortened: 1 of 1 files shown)


lib/election.js
```diff
@@ -771,7 +771,24 @@ function _getManifestFromPeer(
     },
     manifest: callback => api._client.getManifest(
       manifestHash, peerId, callback),
-    sync: ['hashes', 'manifest', (results, callback) => {
+    checkMajority: ['manifest', (results, callback) => {
+      const manifest = results.manifest;
+      const blockHeight = manifest.blockHeight;
+      const manifestItemCount = manifest.item.length;
+      api.getBlockElectors(ledgerNode, blockHeight, (err, result) => {
```
dlongley: I can see us wanting to just pass the `electors` all the way through to here in the future (we do have them but it's several calls up the stack before `_gossipWith`). But... we can leave that to another day to optimize. Perhaps another TODO here above `getBlockElectors` that says: "TODO: consider passing `electors` through to optimize away this look up`"

PR #28, opened 2017-08-25
Verify vote signatures.

### LINKED ISSUE #21: Check result of `jsigs.verify`
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/election.js#L726

TODO: check result

@dlongley needs explanation.

mattcollier:
Fixes #21.

I don't know what the comment // TODO: optimize was about, so I removed it.

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -732,12 +732,15 @@ function _verifyVote(ledgerNode, vote, electionTopic, callback) {
 }
 
 function _verifyVoteSignature(vote, callback) {
-  // TODO: optimize
   jsigs.verify(vote, {}, (err, result) => {
     if(err) {
       return callback(err);
     }
-    // TODO: check result
+    if(!result.verified) {
+      return callback(new BedrockError(
+        'Vote signature verification failed.',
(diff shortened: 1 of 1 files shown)


lib/election.js
```diff
@@ -732,12 +732,15 @@ function _verifyVote(ledgerNode, vote, electionTopic, callback) {
...
-  // TODO: optimize
   jsigs.verify(vote, {}, (err, result) => {
     if(err) {
       return callback(err);
     }
-    // TODO: check result
+    if(!result.verified) {
+      return callback(new BedrockError(
+        'Vote signature could not be verified.',
+        'OperationError', {vote, keyResults: result.keyResults}));
```
dlongley: Hmm, fine other than `OperationError`. I don't think this is an operation error, more like a security error or something. Will think on it more and get back to you.

dlongley: Error text should be:

`Vote signature verification failed.`

With error type:

`AuthenticationError`.

mattcollier:

PR #29, opened 2017-08-25
Add voting round interval.

### LINKED ISSUE #17: Should wait some period of time before communicating again
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/worker.js#L208

TODO: should wait some period of time before communicating again
with the electors to allow gossip/votes to settle?

@dlongley needs explanation.

mattcollier:
Fixes #17.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -21,6 +21,10 @@ cfg.worker = {
   session: {
     // 1 minute
     maxTime: 60 * 1000
+  },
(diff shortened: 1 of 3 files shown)


dlongley:
Let's call it "gossipInterval" instead.

mattcollier:

PR #30, opened 2017-08-28
Validate `status.election`.

### LINKED ISSUE #9: Certify votes
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/worker.js#L309

1. TODO: send votes we know about to peer to optimize
2. TODO: validate `status.election`

Due to a glitch in the reporting tool, I'm not sure which of these two to-dos (or possibly both) we want to address.  Whatever the case, @dlongley, more explanation is needed here.

mattcollier:
Fixes #9.

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -9,6 +9,7 @@ const bedrock = require('bedrock');
 const config = bedrock.config;
 const brLedger = require('bedrock-ledger-node');
 const logger = require('./logger');
+const validate = require('bedrock-validation').validate;
 const BedrockError = bedrock.util.BedrockError;
 
 // load config defaults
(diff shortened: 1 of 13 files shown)


lib/worker.js
```diff
@@ -308,14 +309,16 @@ function _gossipWith(ledgerNode, voter, blockHeight, peer, callback) {
...
-      // TODO: validate `status.election`
       // TODO: make sure `Events` election topic is certified *BEFORE*
       //   roll-call because we depend on the votes for events to exist
       //   in the database (and they only get there if they are certified)
-      return async.eachSeries(status.election, (election, callback) => {
-        api._election.certify(
-          ledgerNode, election.topic, election.electionResults, callback);
-      }, callback);
+      return async.series([
+        callback => validate('continuity.election', status.election, callback),
```
mattcollier: @dlongley it was suggested that by adding validation to the client get block status API, that validation here may be unnecessary.  Is that true?  Does all the election information pass through that API?  Don't we have an issue for adding an API to send election data via a POST handler?

dlongley: Let's just leave it for now. We can look into optimizing it away later.

schemas/continuity.js
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright (c) 2017 Digital Bazaar, Inc. All rights reserved.
+ */
+const schemas = require('bedrock-validation').schemas;
+
+const signature = {
```
dlongley: This exists:

https://github.com/digitalbazaar/bedrock-validation/blob/master/schemas/linkedDataSignature.js

Can we just reuse it? Maybe it needs an option for restricting the linked data signature type to just `LinkedDataSignature2015`. We could potentially just add a TODO to reuse it if that gets us there faster -- and can address later.

mattcollier:
@dlongley revised the signature schema, ready for another pass.

PR #32, opened 2017-08-29
Send votes to peer.

### LINKED ISSUE #15: Send votes we know about to peer to optimize
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/worker.js#L309

TODO: send votes we know about to peer to optimize

@dlongley needs explanation.

mattcollier:
Fixes #15.

This is a trial balloon.  I'm not sure if it's doing the right thing.  But as is, within the test suite, it leads to a minor performance degradation.  Also, it's currently recording the number of successful vote writes on the server side vs duplicate errors.  The number of failures typically exceeds the number of successes by about 30%.  One test suite run: Fail 308/Success 226

@dlongley first, I need you to make sure it's doing the right thing.  Then we would need to discuss under what circumstances there would be a net gain with this optimization so we can try to measure the improvement somehow.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -3,8 +3,10 @@
  *
  * Copyright (c) 2017 Digital Bazaar, Inc. All rights reserved.
  */
+const async = require('async');
 const bedrock = require('bedrock');
 const config = bedrock.config;
+const election = require('./election');
 // NOTE: request.defaults not used here so that request can be stubbed
 const request = require('request');
 const validate = require('bedrock-validation').validate;
@@ -56,25 +58,45 @@ api.getBlockStatus = (blockHeight, peer, callback) => {
 api.getManifest = (manifestHash, peer, callback) => {
   const url = peer + '/manifests?id=' +
     encodeURIComponent(manifestHash);
-  request.get({url, strictSSL, json: true}, (err, res) => {
-    if(err) {
-      return callback(new BedrockError(
-        'Could not get manifest.', 'NetworkError', {manifestHash, peer}, err));
-    }
-    if(res.statusCode !== 200) {
-      return callback(new BedrockError(
-        'Could not get manifest.',
-        res.statusCode === 404 ? 'NotFoundError' : 'NetworkError', {
-          httpStatusCode: res.statusCode,
-          public: true,
-          manifestHash,
-          peer,
-          error: res.body,
-        }));
-    }
-    // TODO: validate manifest
-    callback(null, res.body);
-  });
+  async.auto({
+    get: callback => request.get({url, strictSSL, json: true}, (err, res) => {
+      if(err) {
+        return callback(new BedrockError(
+          'Could not get manifest.', 'NetworkError',
+          {manifestHash, peer}, err));
+      }
+      if(res.statusCode !== 200) {
+        return callback(new BedrockError(
+          'Could not get manifest.',
+          res.statusCode === 404 ? 'NotFoundError' : 'NetworkError', {
+            httpStatusCode: res.statusCode,
+            public: true,
+            manifestHash,
+            peer,
+            error: res.body,
+          }));
+      }
+      callback(null, res.body);
+    }),
+    validate: ['get', (results, callback) =>
+      validate('continuity.manifest', results.get, callback)],
+    validateHash: ['validate', (results, callback) => {
+      const manifest = results.get;
+      const expectedHash = election.createManifestHash(manifest.item);
+      if(manifest.id !== manifestHash) {
+        return callback(new BedrockError(
+          'The hash contained in the manifest is invalid.',
+          'ValidationError', {
+            httpStatusCode: 400,
+            public: true,
+            expectedHash,
+            manifest,
+            peer,
+          }));
+      }
+      callback();
+    }]
+  }, (err, results) => err ? callback(err) : callback(err, results.get));
 };
 
 api.sendEvent = (event, peer, callback) => {
@@ -96,6 +118,28 @@ api.sendEvent = (event, peer, callback) => {
         }));
     }
     callback(null, res.headers.location);
+  });
+};
+
+api.sendManifest = (manifest, peer, callback) => {
+  const url = peer + '/manifests';
+  request.post({url, strictSSL, json: manifest}, (err, res) => {
+    if(err) {
+      return callback(new BedrockError(
+        'Could not send manifest.', 'NetworkError', {manifest, peer}, err));
+    }
+    if(res.statusCode !== 200) {
+      return callback(new BedrockError(
+        'Could not send manifest.',
+        'NetworkError', {
+          httpStatusCode: res.statusCode,
+          public: true,
+          manifest,
+          peer,
+          error: res.body,
+        }));
+    }
+    callback();
   });
 };
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -106,14 +106,16 @@ api.sendVote = (vote, peer, callback) => {
       return callback(new BedrockError(
         'Could not send vote.', 'NetworkError', {vote, peer}, err));
     }
+    // TODO: should the server return success or 409 on duplicate?
     if(res.statusCode !== 201) {
       return callback(new BedrockError(
         'Could not send vote.',
         'NetworkError', {
           httpStatusCode: res.statusCode,
           vote,
-          peer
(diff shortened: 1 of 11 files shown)


lib/client.js
```diff
@@ -92,6 +92,26 @@ api.sendEvent = (event, peer, callback) => {
...
 };
 
+api.sendVote = (vote, peer, callback) => {
+  const url = peer + '/votes';
+  request.post({url, strictSSL, json: vote}, (err, res) => {
+    if(err) {
+      return callback(new BedrockError(
+        'Could not send vote.', 'NetworkError', {vote, peer}, err));
+    }
+    if(res.statusCode !== 201) {
```
dlongley: We should handle some duplicate error code and ignore it here. It's possible that someone else sends the vote before we do.

lib/worker.js
```diff
@@ -325,6 +333,25 @@ function _gossipWith(ledgerNode, voter, blockHeight, peer, callback) {
   });
 }
 
+function _syncVotes(options, callback) {
+  async.each(options.election, (e, callback) => {
+    async.auto({
+      hashes: callback => async.map(e.electionResults, api._hasher, callback),
```
dlongley: We want to avoid recomputing all these hashes (it is very expensive). We already have to do it when calling `certify`. If we don't do `syncVotes` and `certify` in parallel above, perhaps we can figure out a avoid it.

dlongley:
General principles seem correct but I can see how the current implementation may be generating more work than helping.

This is meant to optimize by allowing nodes to determine an election has been won without having to communicate with all of the other nodes, i.e. they can talk to like 10 nodes instead of 100 -- because those 10 nodes already have the election results from the other 90 combined. It's core to the advantage of using a gossip protocol.

mattcollier:
@dlongley renamed to `voteRecord` per your suggestion.

wrt to performance on this patch, is it correct to say that we would not expect to see improvement until we hit scale?

Also, what are your thoughts about whether the server should report success on duplicate votes, or the client should ignore duplicate errors?

dlongley:
> wrt to performance on this patch, is it correct to say that we would not expect to see improvement until we hit scale?

Yes, if there are only a few votes it won't make a big difference. But we want to support many nodes participating in the election.

> Also, what are your thoughts about whether the server should report success on duplicate votes, or the client should ignore duplicate errors?

Well, I was originally thinking having the server return a `409 Conflict` was the way to go, but I think that may be misleading -- since there really is no actual conflict and especially if we want to use that to represent one in the future. So let's have the server return 200 when there is a duplicate. The client should accept either 200 or 201.

mattcollier:
@dlongley on this endpoint a 201 will not be returning a header for the resource correct?  We're just returning 200/201 status code and that's it?

mattcollier:
@dlongley here is an [example vote](https://gist.github.com/mattcollier/1df36a18b2db424769f206f4a94cd058) that is being posted to the server endpoint.

The `topic` is a required piece of information as it is required by the `storage.votes.add` API.  

I note that there is no context on this object.  Should there be one?

Or alternatively, we could post *only* the `vote` to a topic specific endpoint? Such as:
```
https://example.com/consensus/continuity2017/voters/<voterid>/<topic>
```

[Here is the server implementation](http://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/576a59d955cb724cb315bf68d6ed9e5043d59d3b/lib/server.js#L264-L265)

dlongley:
Always 200 (no 201).

mattcollier:
@dlongley syncing of manifest has been added here, ready for another pass.  Please respond with a comment to any new FIXMEs.

lib/election.js
```diff
@@ -818,25 +851,8 @@ function _getManifestFromPeer(
...
-        if(manifestItemCount < twoThirds) {
-          return callback(new BedrockError(
-            'The manifest does not contain the required two-thirds majority.',
-            'ValidationError', {manifest, manifestItemCount, twoThirds}));
-        }
-        callback();
-      });
-    }],
+    checkMajority: ['manifest', (results, callback) =>
+      api.checkMajority(results.manifest, ledgerNode, callback)],
```
dlongley: This should only be called for event vote manifests, not event manifests. Event manifests must have at least 1 event hash -- that's it.

lib/server.js
```diff
@@ -243,6 +242,123 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+              if(result) {
+                return callback(new BedrockError(
+                  'The manifest already exists.', 'DuplicateError', {
+                    manifest: manifest.id
+                  }, err));
+              }
+              callback();
+            })],
+        checkMajority: ['checkExists', (results, callback) =>
+          election.checkMajority(manifest, results.ledgerNode, callback)],
```
dlongley: This check is only correct when the topic is 'RollCall'.

lib/server.js
```diff
@@ -243,6 +242,123 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+        checkItems: ['checkMajority', (results, callback) => {
+          // FIXME: is it necessary to check items if the type is `Events`?
```
dlongley: Yes, we need to check for event hashes, not vote hashes, in the case where the topic is `Events`.

lib/server.js
```diff
@@ -274,7 +390,7 @@ function _createBlockStatus(blockHeight, ledgerNode, voter, callback) {
         if(block.consensusMethod !== 'Continuity2017') {
           return callback(new BedrockError(
             'No Continuity2017 status for block found.',
-            'NotFound', {block, httpStatus: 404, public: true}));
+            'NotFoundError', {block, httpStatus: 404, public: true}));
```
dlongley: Heh, we're going to need to update a lot of legacy code that uses `NotFound` instead of `NotFoundError` if we want to head this way. Thoughts?

lib/server.js
mattcollier: @dlongley to clarify the question, I *believe* it is safe to assume that we already have all the `events`?

dlongley: That's based on a presumption that the client isn't misbehaving -- which we shouldn't assume.

dlongley: Our client won't intentionally misbehave -- but we should not assume that someone else won't write one that will (or that ours won't have bugs).

lib/server.js
mattcollier: I don't know about changing legacy code, but given that `NotFoundError` is being used widely in the ledger code already, I thought a decision had already been made to use `NotFoundError` in new modules.  And I suppose we would continue doing `NotFound` in older modules where it is already be used.

dlongley: @mattcollier, Ok, I didn't realize/remember that we had made that decision. That's fine.

lib/election.js
```diff
@@ -38,6 +38,41 @@ api._getManifest = _getManifest;
...
+  niUri.digest('sha-256', items.join('\n'), true);
+
+/**
+ * Check if a manifest has the required majority vote.
+ *
+ * @param manifest the manifest to check.
+ * @param ledgerNode the ledgerNode that is tracking the manifest.
+ * @param callback(err) called once the operation is complete.
+ */
+api.checkMajority = (manifest, ledgerNode, callback) => {
```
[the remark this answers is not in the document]
mattcollier: @dlongley what if we change the function name, and then do a `NotSupportedError` if the `manifest.type !== 'RollCall'`

lib/election.js
mattcollier: @dlongley does it seem odd to you that I have never observed this code being called in relation to an event manifest?  It's always RollCall, which is why the tests are passing.

lib/election.js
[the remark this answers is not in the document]
dlongley: Then we should name it specifically:

`validateRollCallManifest()`

I do think having the more generic function may be a useful thing in other places too. We can always add it later if you want to make this more specific function now.

lib/election.js
dlongley: I think the test suite isn't constructed so that the nodes ever get different events, right? I think this is something we talked about before and how we need better tests to handle this case.

mattcollier: @dlongley actually, I added a 5th block.  In addition to submitting a unique event at each node, I offset the worker start time which I have found creates a great deal of turbulence (by design).

mattcollier:

PR #35, opened 2017-08-29
Add event validation.

mattcollier:
Added new FIXME comments wrt signing of events.

Depends on: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/30

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -42,9 +42,11 @@ api.getBlockStatus = (blockHeight, peer, callback) => {
         'Could not get block status.',
         res.statusCode === 404 ? 'NotFoundError' : 'NetworkError', {
           httpStatusCode: res.statusCode,
+          public: true,
           blockHeight,
-          peer
-        }, typeof res.body === 'object' || new Error(res.body)));
+          peer,
+          error: res.body,
+        }));
     }
     validate('continuity.blockStatus', res.body, err =>
       callback(err, res.body));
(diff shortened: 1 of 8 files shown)


lib/client.js
```diff
@@ -86,7 +86,8 @@ api.sendEvent = (event, peer, callback) => {
         'NetworkError', {
           httpStatusCode: res.statusCode,
           event,
-          peer
+          peer,
+          error: res.body
         }, typeof res.body === 'object' || new Error(res.body)));
```
dlongley: This `typeof` line is a bit strange, won't this set either `true` or an `Error`? Is that the intent? It seems like it could be fixed up and then we wouldn't need the additional `error` property that's being passed in the details.

mattcollier: @dlongley in addition to the defect your pointed out, this pattern does not work well for two other reasons:
1. It actually does not work at all when doing `new Error`, but does work (as intended) with `new BedrockError`, but of course we can't make the assumption that the object returned adheres to the `BedrockError` API.
2. Dealing with the error like this puts the error details in `error.cause` apart from the other error details in `error.details` which is inconvenient and breaks with our current convention.

I fixed this up to *ONLY* include the error details (res.body) in `error.details` as per our convention.  If we are agreed that this pattern should be abolished, then other instances of its use will need to be removed as well.

mattcollier:
@dlongley would you create the issue related to signature fixme, or tell me what the issue is?

dlongley:
@mattcollier, I just saw a bunch of signature FIXMEs that you added. We should make an issue to address them in the future. A bunch of questions as to whether or not signatures are required.

mattcollier:
@dlongley the fixme wrt signature is relevant to making the test suite work properly, AND whether or not we should be checking for a signature with the validator.

lib/client.js
dlongley: What you did looks good.

mattcollier:

PR #43, opened 2017-08-29
Record `voter.id` hashes for use in sorting.

### LINKED ISSUE #42: Cache hashes to avoid recomputation
http://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/b1a31b0c112f54c85a1206105ec1a92276275dd6/lib/election.js#L159-L160

TODO: cache hashes to avoid recomputation

mattcollier:
Fixes #42.

Need to address the FIXME in the commit.

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -120,7 +120,6 @@ api.getBlockElectors = (ledgerNode, blockHeight, callback) => {
       // it's possible `electors.length` will be less than `electorCount` if
       // a config change happened -- which we allow here; in theory more
       // electors will be recommended later to fill the gap
-
       if(electors.length > electorCount) {
         // TODO: could optimize by only sorting tied electors if helpful
         /*
@@ -155,9 +154,16 @@ api.getBlockElectors = (ledgerNode, blockHeight, callback) => {
             return aPreviousElector - bPreviousElector;
           }
 
-          // 3. sort by hex hash(manifest + voter.id)
-          // TODO: cache hashes to avoid recomputation
-          return _sha256(a.id).localeCompare(_sha256(b.id));
+          // generate and cache hashes
+          // the hash of the previous block is combined with the elector id to
+          // prevent any elector from *always* being sorted to the top
+          a.hash = a.hash || _sha256(
+            results.latestBlock.eventBlock.meta.blockHash + _sha256(a.id));
+          b.hash = b.hash || _sha256(
+            results.latestBlock.eventBlock.meta.blockHash + _sha256(b.id));
+
+          // 3. sort by hash
+          return a.hash.localeCompare(b.hash);
         });
 
         // select first `electorCount` electors


lib/election.js
```diff
@@ -155,9 +157,11 @@ api.getBlockElectors = (ledgerNode, blockHeight, callback) => {
             return aPreviousElector - bPreviousElector;
           }
 
+          // FIXME: does this comment need to be updated?  The existing
+          // implementation does not involve the manifest.
```
dlongley: We do need to involve the manifest or some value from the previous block that changes with every new block and can't be controlled by any particular party. We just want to avoid people creating voter IDs that hash to a low hash in order to give themselves preference.

mattcollier: @dlongley are you saying the comment `3. sort by hex hash(manifest + voter.id)` is accurate, and can be left as is?  And the FIXME at L160 just needs to be removed?

or are you saying we need to alter the inputs to the hash?

dlongley: I'm saying comment we need to alter the inputs to the hash. Either we need to match what comment `3` is saying (if that's easy) or we need to use another input (if that's hard/makes no sense).

mattcollier:

lib/election.js
```diff
@@ -155,9 +154,16 @@ api.getBlockElectors = (ledgerNode, blockHeight, callback) => {
...
           }
 
-          // 3. sort by hex hash(manifest + voter.id)
-          // TODO: cache hashes to avoid recomputation
-          return _sha256(a.id).localeCompare(_sha256(b.id));
+          // generate and cache hashes
+          // the hash of the previous block is combined with the elector id to
+          // prevent any elector from *always* being sorted to the top
+          a.hash = a.hash || _sha256(
+            results.latestBlock.eventBlock.meta.blockHash + a.id);
```
dlongley: Let's do:

```js
a.hash = a.hash || _sha256(results.latestBlock.eventBlock.meta.blockHash + _sha256(a.id));
```

Because it will ensure proper boundaries, etc.

mattcollier: @dlongley Can you say more about why you think this is a good idea?

dlongley: When concatenating different values to produce a hash it's a good idea to make it clear where the boundaries are between the values, particularly if the data is user controlled, so that someone can't manipulate that data to cause confusion. It may be fairly unlikely to happen particularly with our current hash choices for blocks and the fact that the ID comes last. However, I'd rather use a best practice here out of an abundance of caution as the ID can be anything.

It's more of a problem in situations like this:

hash(myName + myFavoriteAnimal)

Where you can have:

myName = alfred
myFavoriteAnimal = fox

And:

myName = alf
myFavoriteAnimal = redfox

mattcollier: @dlongley done, ready for another pass.

PR #45, opened 2017-09-01
Only include votes from the winning round in roll call manifest.

### LINKED ISSUE #44: Duplicate votes being recorded in a block
This issue was discovered while doing integration testing in `bedrock-ledger-agent`, see an example test run on CI: https://ci.digitalbazaar.com/job/bedrock-ledger-agent/40/console

This issue only manifests on the CI system (not on my local system) which likely means it's a race condition.

mattcollier:
Fixes #44.

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -243,8 +243,9 @@ api.certify = (ledgerNode, electionTopic, votes, callback) => {
  * @param ledgerNodeId the ID of the ledger node that is tallying votes.
  * @param blockHeight the height of the block to tally the vote for.
  * @param electors the voting population for the block height.
- * @param electionType the type of votes to tally ('EventOrder' or 'RollCall').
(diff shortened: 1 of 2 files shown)


dlongley:
I'm not sure if it's ok to only include votes from the winning round -- as opposed to *up to and including but not after* the winning round. What are your thoughts on that, @mattcollier?

mattcollier:
@dlongley I believe the answer to your question relates to this: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/34  

Seems like things could get complicated if we start intentionally including votes from N rounds in the block.

Why would we want to record votes for any but the winning round?

dlongley:
Let's proceed with just using the votes from the winning around. I'll have @davidlehn review and see if he can find fault with the approach -- he'll file a new issue if so.

PR #54, opened 2017-09-13
Fix manifest validation.

### LINKED ISSUE #53: New Node Catch-up
Make sure that a newly added node can catch up with the other nodes.

mattcollier:
Fixes #53.

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -38,6 +38,44 @@ api._getManifest = _getManifest;
 api._recommendElectors = _recommendElectors;
 
 /**
+ * Check if a  RollCall manifest has the required number of votes.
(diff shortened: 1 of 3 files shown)


lib/election.js
```diff
@@ -855,6 +883,10 @@ function _getManifestFromPeer(
         return _validateEvents(
           ledgerNode, manifest.item, manifest.blockHeight, callback);
       }
+
+      // FIXME: bypassing this check, votes are in the process of being saved
```
mattcollier: @dlongley I'd like to get this PR merged, and then create another issue for re-implementing the vote check here.

PR #56, opened 2017-09-14
Fix52

mattcollier:
@dlongley failure due to race condition with this patch is:
- local: ~1/30
- aws/jenkins: ~1/60

### DIFF
--- a/lib/election.js
+++ b/lib/election.js
@@ -38,7 +38,7 @@ api._getManifest = _getManifest;
 api._recommendElectors = _recommendElectors;
 
 /**
- * Create a hash of a manifest.
+ * Create a hash of a manifest items (EOL delimited).
  *
  * @param items the items in the manifest.
  * @returns hash the hash of the items.
@@ -1006,14 +1006,13 @@ function _getVoteFromPeer(
   }, callback);
 }
 
+// create manifest hash and store manifest
 function _createManifest(ledgerNodeId, blockHeight, type, items, callback) {
-  // create manifest hash (EOL delimited) and store manifest
-  const hash = niUri.digest('sha-256', items.sort().join('\n'), true);
   const manifest = {
-    id: hash,
-    type: type,
-    blockHeight: blockHeight,
-    item: items
+    blockHeight,
+    id: api.createManifestHash(items),
+    item: items,
+    type
   };
   api._storage.manifests.add(ledgerNodeId, manifest, err => {
     if(err && err.name === 'DuplicateError') {
--- a/lib/election.js
+++ b/lib/election.js
@@ -944,7 +944,15 @@ function _getManifestFromPeer(
             manifest: results.manifest
           }));
       }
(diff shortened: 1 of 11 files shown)


lib/election.js
```diff
@@ -941,9 +993,28 @@ function _getEventFromPeer(ledgerNode, peerId, eventHash, callback) {
   }, callback);
 }
 
+function _getVoteFromPeer(
+  ledgerNodeId, peerId, voteHash, electionTopic, callback) {
+  async.auto({
+    vote: callback => api._client.getVote(voteHash, peerId, callback),
+    hash: ['vote', (results, callback) => api._hasher(results.vote, callback)],
+    store: ['hash', (results, callback) => api._storage.votes.add(
```
mattcollier: @dlongley I believe you commented that we need to do more validation here?

lib/election.js
```diff
@@ -941,9 +993,28 @@ function _getEventFromPeer(ledgerNode, peerId, eventHash, callback) {
...
+        }
+        callback(err);
+      })]
+  }, callback);
+}
+
 function _createManifest(ledgerNodeId, blockHeight, type, items, callback) {
   // create manifest hash (EOL delimited) and store manifest
-  const hash = niUri.digest('sha-256', items.join('\n'), true);
+  const hash = niUri.digest('sha-256', items.sort().join('\n'), true);
```
mattcollier: this should use `api.createManifestHash`

lib/voteStorage.js
```diff
@@ -150,6 +150,24 @@ api.exists = (ledgerNodeId, voteHash, callback) => {
   });
 };
 
+// TODO: add documentation
```
mattcollier: Add documentation.

lib/voteStorage.js
```diff
@@ -175,6 +193,18 @@ api.get = (ledgerNodeId, blockHeight, electionTopic, callback) => {
...
```
mattcollier: Add documentation.

lib/election.js
```diff
@@ -1032,9 +1031,8 @@ function _createManifest(ledgerNodeId, blockHeight, type, items, callback) {
 }
 
 function _createVote(ledgerNode, electionTopic, vote, voter, callback) {
-  // TODO: add logger.verbose() with the vote information here
-  // console.log(voter.id.split('/').pop() + ' votes for ' + electionTopic +
-  //   ', manifestHash=' + vote.manifestHash + ', round=' + vote.voteRound);
+  logger.verbose(voter.id.split('/').pop() + ' votes for ' + electionTopic +
```
dlongley: @mattcollier,

This `voter.id.split('/').pop()` stuff probably needs to change to just `voter.id` (shouldn't assume too much about how these look) ... and we should use the meta data field with the logger and move and/or duplicate some of these items into logged meta data properties. I don't remember how that works, can you take a stab?

mattcollier: @dlongley the api is (message, meta) where meta must be an object.  What do you want the logger message to say?

PR #61, opened 2017-09-15
Check `previousBlockHash` in block status.

### LINKED ISSUE #59: Include `previousBlockHash` in block status and check it when received
We should be including the `previousBlockHash` in the block status data we're sending to clients and those clients should be verifying it and rejecting the block status if it does not match.

dlongley:
Fixes #59.

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -51,16 +51,18 @@ function _sync(session, callback) {
 
   async.auto({
     voter: callback => api._voters.get(ledgerNode.id, callback),
-    blockHeight: callback => _getBlockHeight(ledgerNode, callback),
-    getElectors: ['blockHeight', (results, callback) => {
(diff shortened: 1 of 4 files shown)


lib/worker.js
```diff
@@ -78,12 +80,14 @@ function _sync(session, callback) {
 
 /**
  * Gets the latest consensus block and returns the new proposed block height
- * for the ledger (i.e. the current `blockHeight + 1`).
+ * for the ledger (i.e. the current `blockHeight + 1`) and the latest block
+ * hash as what would become the next `previousBlockHash`.
  *
  * @param ledgerNode the ledger node to get the latest block for.
- * @param callback(err, blockHeight) called once the operation completes.
+ * @param callback(err, {blockHeight, previousBlockHash) called once the
```
davidlehn: Missing closing '}'.

PR #66, opened 2018-02-18
Prevent duplicates in `descendants`.

mattcollier:
Unfortunately set.add does not do a deep compare to prevent duplicates.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has
https://stackoverflow.com/a/29759699/3272112

And since the set wasn't doing anything useful, array.push blows the doors off set.add
https://jsperf.com/set-add-vs-array-push/2

### DIFF
--- a/lib/consensus.js
+++ b/lib/consensus.js
@@ -190,7 +190,7 @@ function _getElectorBranches({history, electors}) {
 function _getAncestorHashes({allXs, ledgerNodeId}) {
   // get all ancestor hashes from every consensus X
   const hashes = new Set();
-  const descendants = new Set();
+  const descendants = [];
   const parentHashes = new Set();
   for(const x of allXs) {
     // TODO: anything missed or different here with byzantine forks?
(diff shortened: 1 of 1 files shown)


dlongley:
> Unfortunately set.add does not do a deep compare to prevent duplicates.

Does it need to? I think the code here is only comparing object references, with the exception of regular events.

In any case, this change causes it to fail the tests. I'll leave a specific comment on the line that causes the trouble.

lib/consensus.js
```diff
@@ -211,9 +211,10 @@ function _getAncestorHashes({allXs, ledgerNodeId}) {
           if(!hashes.has(parentHash) &&
             !event._parents.some(p => p.eventHash === parentHash)) {
             parentHashes.add(parentHash);
+            hashes.add(parentHash);
```
dlongley: This line changes how the algorithm works and causes regular events to not have their ancestry processed and added.

dlongley: I think I have a fix for it... working on it.

mattcollier:
`orders` is full of duplicates.  If using `hashes` is a problem, a new set can be created.

dlongley:
Just pushed a fix, what do you think? Tests pass.

PR #69, opened 2018-02-24
Add additional coolDown to gossip cycle.

mattcollier:


lib/agents/gossip-agent.js
```diff
@@ -118,12 +118,19 @@ module.exports = class GossipAgent extends ContinuityAgent {
...
         // start the cycle over
         return callback();
       }
-      callback(err);
+      if(err) {
+        return callback(err);
+      }
+      if(!(results.notification && results.coolDown && this.halt)) {
+        const {coolDownPeriod} = config['ledger-consensus-continuity'].gossip;
+        setTimeout(() => callback(), coolDownPeriod);
```
dlongley: Add a return.

PR #72, opened 2018-03-17
Block add

mattcollier:
Depends on https://github.com/digitalbazaar/bedrock-ledger-storage-mongodb/pull/16

### DIFF
--- a/lib/blocks.js
+++ b/lib/blocks.js
@@ -12,6 +12,7 @@ const bedrock = require('bedrock');
 const cache = require('bedrock-redis');
 const {config} = bedrock;
 const database = require('bedrock-mongodb');
+const hasher = require('bedrock-ledger-node').consensus._hasher;
 // const logger = require('./logger');
 const {BedrockError} = bedrock.util;
 
@@ -187,6 +188,18 @@ api.write = ({ledgerNode, state, consensusResult}, callback) => {
       }, callback)],
   }, err => callback(err, err ? false : true));
 };
+
+api.writeGenesis = ({block, ledgerNode}, callback) => async.auto({
+  hashBlock: callback => hasher(block, callback),
+  writeBlock: ['hashBlock', (results, callback) => {
+    const meta = {
+      blockHash: results.hashBlock,
+      consensus: true,
+      consensusDate: Date.now()
+    };
+    ledgerNode.storage.blocks.add({block, meta}, callback);
+  }]
+}, callback);
 
 // TODO: proofs should be stored and retrieved in the same fashion as events
 function _expandConsensusProofEvents({block, ledgerNode}, callback) {
--- a/lib/blocks.js
+++ b/lib/blocks.js
@@ -47,20 +47,49 @@ api.getNextBlockInfo = (ledgerNode, callback) => {
     if(err) {
       return callback(err);
     }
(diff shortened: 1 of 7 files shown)


lib/blocks.js
```diff
@@ -42,29 +46,68 @@ api.getNextBlockInfo = (ledgerNode, callback) => {
...
+          return callback();
+        }
+        // this should never happen
+        if((blockHeight - cachedBlockHeight) !== 1) {
+          return callback(new BedrockError(
+            'Cache is behind by more than one block.', 'InvalidStateError',
+            {blockHeight, cachedBlockHeight, ledgerNodeId}));
+        }
+        // inconsistency needs to be repaired
+        _repairCache({blockHeight, ledgerNode}, callback);
```
dlongley: Why are we calling the repair code within `getLatest`? This call can be called from anywhere -- right? If so, is it necessary to do this cache repair when that happens (i.e. is the cache even used in that case?) How safe is it to have two processes make this call at once (i.e. outside of the consensus worker lock)?

Would it be better to just call the repair code once at the start of the consensus worker (or beginning of or start of loop of whatever agent is using the cache)?

lib/blocks.js
```diff
@@ -126,29 +189,55 @@ api.write = ({ledgerNode, state, consensusResult}, callback) => {
...
+        const now = Date.now();
         callback(null, {
           block: block,
-          meta: {blockHash}
+          meta: {blockHash, consensus: true, consensusDate: now}
         });
       });
     }],
-    store: ['createBlock', (results, callback) =>
+    updateKey: ['createBlock', (results, callback) => {
```
dlongley: Do we have a possible inconsistency issue with updating keys here as well?

If the block is written but we fail to update keys ... will that be a problem? I think the old "incomplete block" code used to prevent that somehow so I just want to double check that it's not an issue now with these changes. The repair cache code could also re-execute the update keys code as a possible fix (if needed).

mattcollier: It is no longer possible for the block to be written without the keys being updated.  

https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/41e96f26fd6d6a30ebc1a81079901e5c8f3046de/lib/blocks.js#L189-L208

lib/blocks.js
mattcollier: Relocated the consistency check.

lib/blocks.js
dlongley: If we fail after the keys are updated and later try to write the block again, will it have the same keys/key references in it as before?

lib/blocks.js
```diff
@@ -123,16 +125,19 @@ api.write = ({ledgerNode, state, consensusResult}, callback) => {
...
-            {id: key.id, seeAlso: key.seeAlso} : {
-              id: key.id,
-              type: key.type,
-              owner: key.owner,
-              publicKeyBase58: key.publicKeyBase58
-            });
+          // `seeAlso` may have been updated during a prior failed attempt to
+          // write the block
+          const publicKey = result.map(key =>
+            (key.seeAlso && key.seeAlso !== blockId) ?
```
dlongley: Great!

lib/blocks.js
mattcollier: Good catch.  I added a provision to address this situation.

PR #73, opened 2018-03-21
Merge event transaction

mattcollier:
This follows same pattern as blocks.add
- writing merge event is a single mongo insert operation, if it fails, the merge may be repeated again later
- if the cache update fails or never executes, this condition is detect in the merge-agent and the cache is repaired.


lib/events.js
```diff
@@ -1061,7 +920,10 @@ api.getHeads = ({eventHashes, ledgerNode}, callback) => {
 // this is not just for local branch, this finds the latest merge event from
 // the specified creator
 api._getLocalBranchHead = (
-  {creatorId, latest = false, ledgerNode}, callback) => {
+  {useCache = true, creatorId, latest = false, ledgerNode}, callback) => {
+  if(!useCache && latest) {
+    throw new Error('`cache` and `latest` arguments must both be true.');
```
dlongley: `cache` => `useCache`

PR #79, opened 2018-06-08
 Add hashed recordId to operations

mattcollier:
Also refactored event/operation validation.  For peer events, it needed to be moved forward in the processing procedure.

Also did some code cleanup.


lib/events.js
```diff
@@ -479,22 +479,6 @@ api.create = ({ledgerNode}, callback) => {
   });
 };
 
-api.headDifference = ({creatorHeads, eventsCollection}, callback) => {
```
dlongley: I'm assuming this was ok to remove.

mattcollier:
Test suite is failing on CI due to changes in bedrock-ledgers-storage APIs, working on that now.

PR #95, opened 2018-08-20
Multihash

mattcollier:
This may still be WIP, but creating PR for tracking.

Test suite is passing on node 10.9.0 on commit 3ecc362e4d670c74815be4bba391e03c30e4903b

PR #96, opened 2018-08-20
Config event

mattcollier:
This depends on #95 

Prior to rebasing on `multihash`, this test suite had several thousand passes on testcloud.

It is now based on multihash and testcloud has just been launched.  This will be a good test of the `multihash` stack.

We will do the same withe the `promises` stack when it is complete.


mattcollier:
Replaced by: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/102

PR #110, opened 2018-08-24
Remove public key cache

dlongley:
Tests passing, ready for review.

PR #111, opened 2018-08-26
Cache electors

dlongley:
Needs `remove-public-key-cache` PR merged first.


dlongley:
This is ready for review, but review will be easier once PR #110 is in.

PR #114, opened 2018-08-26
Remove eventHash from the root of cached merge events.

mattcollier:


lib/cache.js
```diff
@@ -185,14 +185,10 @@ api.events.addLocalMergeEvent = async ({event, meta, ledgerNodeId}) => {
   // processing peer merge events
   // TODO: `creator` is quite a long URL, can a substitution be made?
 
-  // for local merge events we only cache a summary of the event; it does not
-  // include its operations and it puts `eventHash` at the top level to
-  // assist the consensus algorithm (it is not put in `meta` like it is
-  // for persistent ledger storage (e.g. mongo)
+  // for local merge events, only cache a summary of the event
```
dlongley: I recommend changing the comment to something that explains why we're only doing a summary:

```for local merge events, only cache a summary of the event because that is all that is needed for consensus to be computed```

PR #125, opened 2018-08-29
Use filterHashes API in getAncestors.

mattcollier:
Depends on: https://github.com/digitalbazaar/bedrock-ledger-storage-mongodb/pull/38


lib/election.js
```diff
@@ -198,35 +198,34 @@ async function _getAncestors({blockHeight, hashes, ledgerNode}) {
...
+  const nonConsensusSet = new Set(nonConsensusHashes);
+  const notFound = parentHashes.filter(h => !nonConsensusSet.has(h));
+  // the test suite does not pass `blockHeight` into this API
+  if(notFound.length !== 0 && _.isInteger(blockHeight)) {
+    const hashes = await filterHashes({blockHeight, eventHashes: notFound});
+    nonConsensusHashes.push(...hashes);
   }
   return hashes.mergeEventHashes.concat(nonConsensusHashes);
+
+  async function filterHashes({blockHeight, consensus, eventHashes}) {
```
dlongley: This makes me wonder if we should have a `batch` option at the storage layer.

mattcollier:
@dlongley this needs another round.  The calculatedDebounce code that was removed from the merge agent was abandoned many moons ago, but I put it in the phantom zone: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/tree/old-merge-agent-debounce

lib/blocks.js
```diff
@@ -10,20 +10,10 @@ const bedrock = require('bedrock');
...
 module.exports = api;
 
-// TODO: document
-api.getLatest = callbackify(async (ledgerNode) => {
-  const latest = await ledgerNode.storage.blocks.getLatest();
-  // mutates block
-  await _expandConsensusProofEvents(
-    {block: latest.eventBlock.block, ledgerNode});
-  return latest;
-});
```
dlongley: I went looking for this over here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity-storage ... but didn't see it. What's the deal with this change?

lib/election.js
```diff
@@ -198,35 +198,34 @@ async function _getAncestors({blockHeight, hashes, ledgerNode}) {
       'Some ancestors selected for consensus do not exist.',
       'InvalidStateError', {ledgerNodeId, parentHashes});
   }
+  const nonConsensusHashes = await filterHashes(
+    {consensus: false, eventHashes: parentHashes});
 
-  // retrieve up to 1000 at a time to prevent hitting limits or starving
-  // resources
```
dlongley: I take it that this concern has been mitigated?

lib/blocks.js
dlongley: @mattcollier,

I think we failed to actually compress those. We need to do that ... so we probably don't want to remove that code.

dlongley: I think, right now, we're just storing *entire* consensus proof events in the block. We should confirm and fix that in another PR, if so.

mattcollier: A block in the database looks like below.  This `getParticipants` API is what is being used now to unwind the proofs: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/d13228f585e87f548388a4b9fb3905e7531f2cf7/lib/blocks.js#L27-L28
```
{
    "_id": ObjectID("5b92d7af1087bd06440b9e01"),
    "block": {
        "@context": "https://w3id.org/webledger/v1",
        "id": "did:ledgertest:eb8c22dc-bde6-4315-92e2-59bd3f3c7d59/blocks/9",
        "blockHeight": 9,
        "consensusMethod": "Continuity2017",
        "type": "WebLedgerEventBlock",
        "consensusProofHash": [
            "zQmd1XN8JzHScXPbNcPabrJahVQwn9Hgb8ekKYPcsJaqCFV",
            "zQmNZaX7USPXvTSjaos4PjKrQAJ4a9JxfYDSunx1gZA2Cxb",
            "zQmQbhY36EwKgAuBKQptnwze1bz6U81BE3KbtZBBPQhyEdY",
            "zQmVaZxgsEvjTKnMfg1Qtsh4Hbd1MBHLZFkodkHKsc9QJKa",
            "zQmT35QxXkxjcEzSHSfH1SLu9WZhqGDMzfdCZesPB6YDz74",
            "zQmVdhshAP5s5jTvBhrD48nUUKvjAxgmN6aSLcagNtR2Zdt",
            "zQmeFjexzXveMi5nsLQmSgNPU3F9XQnq1r8ZkU7t2PoNfAg",
            "zQmQYSi9dCwcAuDYygps3oksngBEoziRuQLsH1YfCtmoaJE",
            "zQmZvj8BPzTWxr5ChNeAsF78BJXq6TeEd5GgyXxEN4WbYrE",
            "zQmXysR6vLNuhesa52CFLXWKwCuNG46nHU1KQhEDnkazMZH",
            "zQmXnCtrrkUyZyvh45kpGPX3kCdahnaq7oikMriaqeHTgkD",
            "zQmSy52NpPXxJY7wV8myt5F6jGKmrpb4zu53dkPdNcwjrLk"
        ],
        "previousBlock": "did:ledgertest:eb8c22dc-bde6-4315-92e2-59bd3f3c7d59/blocks/8",
        "previousBlockHash": "zQmRsLikBs3t8DzrtmsRBFj4Tzqo7Z7dKEgb5cPr8sGE1GM"
    },
    "id": "oQBZy+yRqihvz2cX5ZQ4w5UrbGXOskuvEK+KphGuIjc=",
    "meta": {
        "blockHash": "zQmbCmA7xFyQtBgT5SThY1QpwrA26H4NzcaHSSYZtBtAumd",
        "consensus": true,
        "consensusDate": 1536350127842,
        "created": 1536350127844,
        "updated": 1536350127844
    }
}
```

dlongley: That probably means that if someone were to get a "full" block from storage they wouldn't get all the consensus proof events in it... which is not correct. In the same way we expand events in a block we should expand the consensus proof events. That's what that other code was for ... but I guess it stopped getting called somewhere along the way.

PR #126, opened 2018-08-30
Add support for storage plugins.

mattcollier:


dlongley:
@mattcollier,

How does this extra layer of abstraction help us?

PR #127, opened 2018-09-10
Redis gossip / Ledger Configuration Change / Cache Refactor / GossipPeerSelector

mattcollier:
The PR contains the following changes:
- all events for gossip come directly out of redis cache (without a JSON.parse) and are streamed to the peer.  A mongo $graphLookup is still used to determined which events need to be communicated.
- Refactored the cache APIs into separate files
- Implements GossipPeerSelector and GossipPeer for selecting gossip peers and tracking communication issues with peers. In brief, peers that cannot be reached start with a backoff of 5s which is multiplied by a configurable `factor` (default = 2) on each subsequent failure.  A successful contact resets the backoff to zero. The list of candidate peers is limited to those peers that participated in some way in the most recent block.
- Stops all Continuity agents when a `WebLedgerConfigurationEvent` reaches consensus.  This causes the consensus work session to terminate, and a new `LedgerNode` instance will be created before the next work session begins that utilizes the new ledger configuration. There are new unit tests demonstrating that this mechanism can be used to load a new consensus algorithm.
- Replaces the pure js JSONStream parser with an implementation that uses JSON.parse.

1.3M events and 75M operations passed through this code over the weekend.


lib/agents/merge-agent.js
```diff
@@ -91,21 +92,35 @@ module.exports = class MergeAgent extends ContinuityAgent {
...
-        this.gossipAgent.sendNotification(err => {
+        // there is a new merge event, or the last notification attempt
+        // did not result in a gossip session
+        // it is imperative that notifications eventually result in a gossip
+        // session in order for the local node to have its events included
+        // in the ledger
+        callbackify(async () => {
+          return Promise.all([
+            await _cache.gossip.notifyFlag({add: true, ledgerNodeId}),
+            await this.gossipAgent.sendNotification()
```
dlongley: Should these `await`s be here?

lib/blocks.js
```diff
@@ -157,17 +157,15 @@ api.write = callbackify(async ({ledgerNode, state, consensusResult}) => {
     meta: {blockHash, consensus: true, consensusDate: now}
   };
 
-  // FIXME: the events must be checked in the block -- if any of them is
-  //   a configuration event, then further processing in the worker should
-  //   halt, allowing for a different worker code path to run after the
-  //   new configuration has been validated and accepted (or rejected)
```
dlongley: We actually implemented what this FIXME says now, right? Can we capture this comment (with whatever tweaks are necessary) in the code documentation somewhere?

lib/server.js
```diff
@@ -67,39 +69,74 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
       const voterId = config.server.baseUri +
         '/consensus/continuity2017/voters/' + req.params.voterId;
       // FIXME: place limits on size of eventHash
       let ledgerNodeId;
-      let ledgerNode;
-      const eventHashes = req.body.eventHash;
+      // let ledgerNode;
       try {
         ledgerNodeId = await _voters.getLedgerNodeId(voterId);
+        // ledgerNode = await brLedgerNode.get(null, ledgerNodeId);
```
dlongley: Remove now unused commented code?

lib/agents/merge-agent.js
mattcollier: fixed.

PR #131, opened 2018-09-18
Recovery mode electors

dlongley:
This still needs tests and a flag in the consensus-agent for tracking `recoveryMode`. Otherwise, ready to go.

Tests should include taking one or more of the `alpha`/`beta`/etc histories and changing them so that two of the four electors do not communicate with the other two for enough merge events, such that the `recoveryGenerationThreshold` is passed and the recovery mode electors take over and make a new block.


mattcollier:
@dlongley I think the test method would be something like this:

1. start with one of the 4 node X block frameworks
2. run with 4 nodes for a number of blocks
3. exclude 2 of the 4 nodes from the worker cycle
4. after the expected number of worker cycles on the active nodes, a block should be created. and we can make assertions about the consensus proof etc.

Sound reasonable?

dlongley:
@mattcollier,

Yes. That's the simplest test. The complication may be in "choosing" which ones are the recovery ones... if you can see what the elector selector plugin returns before you start the worker cycle, then you're good, it will tell you what it picked for the recovery electors and you can just turn the other two off.

Another test that would be nice to have would be a "no fork" test ... which would be harder to engineer ... and we might need to use the alpha/beta history stuff for. Essentially, we want to get *just* to the point where a block is going to be made and then don't talk to the node we need the event from to make it until the `recoveryGenerationThreshold` passes ... and then talk to the node. Then ALL of the nodes reach consensus WITHOUT triggering recovery mode. A second test can do the same thing with one change ... let the `recoveryGenerationThreshold` pass and let the recovery electors talk *after* that ... and then let the consensus event pass ... here ALL of the nodes should enter recovery mode.

dlongley:
@mattcollier,

Please see my edit above if you already read my message. Also, I left the `flag` off in the consensus agent because I didn't know how you'd want that to be implemented. It's extremely simple code ... but was going to let you do it as you know how that thing works. Do the TODOs make sense?

dlongley:
Fixed a few things and ran some manual tests. Done with this for now... @mattcollier, ready for you to add the aforementioned flag and tests.

mattcollier:
@dlongley Yeeaaah, I'm going to need you to dive back into this.  I've added a test with the singular objective to trip the recovery mode error being thrown, which is not happening.  See:
https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/commit/e38eda6a744c57e5f0cb00020ab10cfafc98687d

First, we need to ensure that the test is creating the proper conditions.  The test is described here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/2c6943b7c8e9bcace2a9b93e0e00ca785c993af7/test/mocha/100-recovery-mode.js#L146-L163

The test will currently run indefinitely without throwing a `NewElectorsRequiredError`.

I've confirmed that a fresh install of the test suite produces consistent results.

The two observations I would make is that `DECISION COUNT` does not rise above 1 which appears to be because `DECISION` and `PROPOSAL` are always `undefined`

Even during the part of the test where 7 nodes are functioning normally, it seems to be falling into some portion of the recovery mode loop.  Perhaps this is expected.

dlongley:
@mattcollier,

You want to see "NEW ELECTORS SEEN" to be zero a bunch ... otherwise it's making progress.

mattcollier:
@dlongley with latest commit b113c23, the test is now starting to log  'NEW ELECTORS SEEN' after the 3 recovery electors cross the threshold.  It seems to do this indefinitely with values like 0, 1, 2.  The `NewElectorsRequiredError` is never returned.

This requires installing the latest elector selection code.

https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/b113c23bba673b9804f6e4087c8a1889092e2588/lib/consensus.js#L1151-L1152

dlongley:
> Is that what you are looking for?

Well, that's good that it passes, but I'm looking for a minimal test that fails. So we'll need to make another test that actually fails, triggering the bug we're seeing in the other more complex test.

mattcollier:
@dlongley it appears by increasing the number of starting nodes to 7 (1e1d6d4), and deleting four of those, I have been able to replicate what we were seeing in the test last week.  It seems to be waiting on a proposal that never comes.

mattcollier:
For the record, ~4K successful passes on d5bf8070777927c61ad912da88187a5d6cc67177

PR #137, opened 2018-12-03
Make the setting of stricSSL evented.

mattcollier:


mattcollier:
OK, we'll lose the event and use config vars.

PR #139, opened 2018-12-05
Basis block height

mattcollier:


dlongley:
Awesome, thanks @mattcollier!

PR #141, opened 2018-12-18
Async agents

mattcollier:
This is based on the `eventCreator` branch.

Make all consensus agents use async/await.  There are no longer `async` or `request` deps in this module!

## TestCloud
commit 7db8c1e 2700 passes #asyncAgents

Putting the last little bit on TestCloud now.


mattcollier:
400 TestCloud passes on commit d0afca4. Merging this.

PR #143, opened 2019-01-22
Jsigs3

mattcollier:
@dlongley a couple fixmes in signature.js to go over.


mattcollier:

mattcollier:
@dlongley what is the trigger that causes jsigs to put `validationMethod` in the proof instead of `creator`?

dlongley:
> @dlongley what is the trigger that causes jsigs to put validationMethod in the proof instead of creator?

I think it's if the `key` passed to the signature suite has a `controller` property or not. We should probably change it so it only uses `creator`` if the key has an `owner` property (which is what I thought it was doing, but I guess not).

dlongley:
https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/suites/JwsLinkedDataSignature.js#L45-L50

^Needs "controller" to exist to use new `verificationMethod`. I'll file an issue to change to what I said above.

mattcollier:

Also I want to make sure the merge event proof is proper, the `proofPurpose === 'assertionMethod'` seems suspicious.
```
{
  "type": "Ed25519Signature2018",
  "created": "2019-01-22T21:06:59Z",
  "verificationMethod": "https://bedrock.localhost:18443/consensus/continuity2017/voters/z6MkmUTeg2BuKcPUSQkJbKmyEpNNif9d45GzKGUp2fXszZZZ",
  "proofPurpose": "assertionMethod",
  "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..byWqCQMlfst8ZdEk35Vi2XV4UR4O1uKGx_dilfq7g891eORclkAqIlKmADtCpNp1aKqU9E6yWLw6WSSPNweyBQ"
}
```

PR #146, opened 2019-02-13
Use jsonld-signatures@4.

mattcollier:
As anticipated, no changes needed here related to this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "body-parser": "^1.18.2",
     "bs58": "^4.0.1",
     "crypto-ld": "^3.0.0",
-    "jsonld-signatures": "^3.3.0",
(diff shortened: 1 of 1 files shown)


dlongley:
@mattcollier, don't we have to set `useNativeCanonize` flags in several places?

mattcollier:
@dlongley wrt useNativeCanonize it is my intention to eventually do some testing to see if that feature is still beneficial in light of the latest benchmarking findings.  So my proposal would be to launch testnet v2 without it since probably won't have time do to thorough testing before that.  We will however do some basic performance tests using JS canonize code.  If we see problems there, then perhaps we would want to investigate this fully sooner rather than later.  How do you feel about that idea?

dlongley:
@mattcollier, let's proceed as you suggest.

PR #147, opened 2019-02-15
Add genesis merge exception in validateGraph.

mattcollier:


lib/events.js
```diff
@@ -1032,8 +1033,22 @@ async function _validateGraph({eventHashes, eventMap, ledgerNode}) {
...
+                'DataError', {
+                  ancestorRecord,
+                  directAncestor,
+                  eventRecord,
+                  httpStatusCode: 400,
+                  public: true,
+                });
+            }
+            // a creator mismatch is acceptable if the event is based on the
+            // generation 0 merge event (genesis merge event)
```
dlongley: The comment should probably read the opposite since the condition is checking for an error ... so more like "creators must match after the genesis merge event (after generation 0)"

PR #149, opened 2019-02-17
Linter

mattcollier:


lib/consensus.js
```diff
@@ -3,6 +3,7 @@
  *
  * Copyright (c) 2017-2018 Digital Bazaar, Inc. All rights reserved.
  */
+/* eslint-disable no-unused-vars */
```
davidlehn: Too many warnings without this?

lib/voters.js
```diff
@@ -110,7 +110,7 @@ api.get = callbackify(async (
   return voter;
 });
 
-api.getLedgerNodeId = callbackify(async (voterId) => {
+api.getLedgerNodeId = callbackify(async voterId => {
```
dlongley: I guess we still need these `callbackify` calls because of the test suite? +1 for removing them as soon as we can.

lib/consensus.js
mattcollier: Correct.

lib/voters.js
mattcollier: Agreed.

PR #157, opened 2019-05-20
Request needed events in chunks.

mattcollier:
Somewhat recently a limit was added on the HTTP gossip endpoint for events that limit the number of events that can be requested at one time (250).  This PR adds the corresponding limits on the requesting side by separating the needed list of events into multiple chunks.


dlongley:
Readability could potentially be improved by moving the stuff inside the for loop into a separate function that grabs a single chunk.

PR #161, opened 2019-08-21
Fix batch gossip

mattcollier:


lib/agents/gossip-agent.js
```diff
@@ -42,18 +43,48 @@ module.exports = class GossipAgent extends ContinuityAgent {
...
+          do {
+            const {eventBlock: {block: {blockHeight}}} =
+              await ledgerNode.blocks.getLatestSummary();
+            ({deferredEvents, processedEventHashes} = await _events.addBatch({
+              blockHeight, events: deferredEvents, eventMap, ledgerNode,
+              processedEventHashes
+            }));
+            if(deferredEvents.length > 0) {
+              // if events were deferred, wait for a new block and retry
+              await delay(250);
```
mattcollier: If we were to have a listener for block creation, it would go here, but I'm not sure there's much benefit to that complexity.  We expect that a short time after `addBatch` on L73 completes one or more new blocks will be created.  I believe it's possible that an add block event may have already fired by this point so we would likely have to do some promiseHelperLib.race() thing waiting on listener or some hard timeout anyway.

mattcollier:
I should add that we have a node catch-up test here that was instrumental in getting this working: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/8ef1da6f69866654a9c04669ff7ae51b979a9404/test/mocha/090-multinode.js#L341-L342

This test models exactly the issue we are are attempting to address here.

lib/gossip.js
```diff
@@ -191,14 +190,9 @@ async function _getNeeded({ledgerNode, needed, peerId}) {
...
-          // the correct number of events have been received, submit the batch
+          // the correct number of events have been received
           if(events.length === needed.length) {
-            return _events.addBatch({events, ledgerNode, needed})
-              .then(resolve)
-              .catch(error => {
-                logger.error('An error occurred in `addBatch`', {error});
-                reject(error);
-              });
+            return resolve();
```
dlongley: Seems like this could be combined with the above conditional now -- they both do the same thing.

lib/agents/gossip-agent.js
dlongley: Ok, we might just want to leave a comment about considering doing some kind of push notification in the future but that there might be race conditions that make it complicated. Nothing to do right now.

mattcollier:
Thanks @dlongley implemented all your suggestions.

mattcollier:
@dlongley just added another commit with at check for deferred events which avoids a bunch of looping in the common case.  Can you check my logic on that?  Tests are passing, and I will be putting this on testcloud before merging.

lib/events.js
```diff
@@ -153,7 +153,7 @@ api.addBatch = async ({
         throw result.error;
       }
     } else if(jsonld.hasValue(event, 'type', 'ContinuityMergeEvent') &&
-      _temp.insideBatch.length !== 0) {
+      deferredEvents.length > 0 && _temp.insideBatch.length !== 0) {
```
dlongley: Seems like we should use `_temp.insideBatch.length > 0` for consistency with the other length check.

That said, can you explain why this is ok to skip when there are no deferred events? Is this saying that if there are no deferred events then the merge event can't possibly be a descendant of anything that was deferred, so there's no reason to consider deferring it as well? If so, and the ordering that events are visited here ensures that merge events are always visited *after* their ancestors, then this check seems fine to me. I'm assuming the order is *trusted* (we aren't relying on an untrusted peer's ordering of the events we're processing, but rather, we've made sure it's legit).

mattcollier: Yes, you have the right idea.  Ordering of gossip is currently occurring on the gossip target side of things using the kahn's algo.  Since we're are validating the DAG, we know that everything is in proper order.  So there is really not a trust issue here, it's just a matter of defining the protocol.  If the peer did not order things properly, the gossip session will be aborted and all the disincentives for doing things incorrectly on the network apply (back-off etc). 

As part of this PR, we are now hashing all the events as a preliminary step.  So it would be easy to add a sort on the receiving side of things instead of or in addition to the target peer side.  This needs further consideration and if we decide to do something differently, it would need to be in a separate PR.

mattcollier:
In addition to the test suite passing, This passed 20 iterations of 4 node X 100 op/sec, 20 min ledger deployments on TestCloud battery: d52839c9-3ccc-47bb-9318-8c7412dd3623

PR #162, opened 2019-08-26
Implement maxRetries on deferred events.

mattcollier:


mattcollier:
@dlongley hopefully last pass, then it can go onto TestCloud.

PR #165, opened 2019-10-01
Decide when seeing `f+1` endorsed proposals for the same set.

dlongley:
This is an experimental branch with an optimization for decisions. Instead of waiting for `2f+1` endorsed proposals, only `f+1` are necessary as they implicitly hide safe supermajority support in them. We need to run some tests on this, but a preliminary proof supports this concept.


mattcollier:
This was merged to 4.x branch.

PR #179, opened 2019-11-15
Update ContinuityAgent.start() to use Promises.

davidlehn:


test/mocha/060-worker-gossip-with.js
```diff
@@ -256,7 +256,9 @@ describe.skip('Worker - _gossipWith', () => {
           {consensusApi, ledgerNode: nodes.alpha, mockData}, callback)],
       writer: ['remoteEvents', (results, callback) => {
         const eventWriter = new EventWriter({ledgerNode: nodes.alpha});
-        eventWriter.start(callback);
+        eventWriter.start()
```
mattcollier: might be doing a lot of this... how about:
```
callbackify(eventWriter.start)(callback);
```

dlongley: I also don't think what's currently here is exactly the same as it was before ... if an error is thrown by the `callback` it would fall through to the `catch` (which it wouldn't have done in the past). Instead, this would be equivalent (I think):

`eventWriter.start().then(() => callback(), callback);`

davidlehn: I think callbackify on methods is awkward.  Need to do weirdness like `callbackify(eventWriter.start.bind(eventWriter))(callback)` or similar.  I figured using `then()` was easier to read, and easier to convert to async/await if that happens.

For now I'll fix the `then()` usage to the two param type.

PR #181, opened 2019-12-05
Update ledger configuration schema.

mattcollier:


mattcollier:
@dlongley looking for an approval here.

PR #182, opened 2019-12-07
Implement history.getRecentHistory API.

mattcollier:
@dlongley here's the API you requested.  It's my suggestion that the gossip and event writer agent be allowed to operate basically as it is now, but with better guidance on peer selection.

Flushing the event writer immediately before a merge or computing a merge would be great.

As for the merge and consensus agents, maybe those can be pipelined together or what have you.

All tests are passing with this except for one assertion as noted in a fixme.

I structured this so it could easily be cherry picked onto another branch.

### DIFF
--- a/lib/history.js
+++ b/lib/history.js
@@ -1,0 +1,148 @@
+/*!
+ * Copyright (c) 2017-2019 Digital Bazaar, Inc. All rights reserved.
+ */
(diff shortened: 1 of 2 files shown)


dlongley:
Closing as we've moved in a different direction.

PR #185, opened 2019-12-14
Single pipeline

mattcollier:


lib/consensus.js
```diff
@@ -2211,31 +2083,190 @@ function _computeEndorsement({
...
+      if((!_hasEndorsedProposal(mra) && _hasEndorsedProposal(otherHead)) ||
+        (!_hasEndorsedRecoveryProposal(mra) &&
+          _hasEndorsedRecoveryProposal(otherHead))) {
+        priorityPeers.add(elector);
+        break;
+      }
+
+      // if `otherHead`'s support has switched, then prioritize `elector` to
+      // merge it in
+      if((!mra && otherHead._supporting) ||
```
mattcollier: the `!mra` case has already been handled by the conditional on L2173.

lib/consensus.js
```diff
@@ -2211,31 +2083,190 @@ function _computeEndorsement({
...
+        (mra && !_compareSupportSet(mra._supporting, otherHead._supporting))) {
```
mattcollier: `mra` will always be true here because the `!mra` case has already been handled by the conditional on L2173.

PR #187, opened 2019-12-17
Fix multihashes dependency.

davidlehn:
Should be using 'multihashes' instead of 'multihash'.


davidlehn:
bedrock-ledger-node pulls it in, so the require worked.

PR #189, opened 2019-12-18
Implement rate limiting of operations.

mattcollier:


mattcollier:
@dlongley I mentioned elsewhere that I don't think this implementation would help under the conditions we discussed yesterday.  That is the inspiration behind getting stats integrated into prometheus.  We need some visibility into the ledger when the other disk latencies etc kick in to determine what metric we might use to help there.

With the existing gossip implementation, I don't see how it's possible that there would be a large number of outstanding merge events and thus would not have triggered the cutoff.  We need to determine what ledger metrics *do* have a correspondence with disk latency issues.

dlongley:
@mattcollier -- ok, makes sense.

dlongley:
This has been overtaken by events and we have other mechanisms for limiting ops, closing.

PR #194, opened 2020-02-19
Bug fix for permit calculation.

dlongley:
Making a PR for this before we forget about it.


dlongley:
Nevermind, it turns out this branch is 80 commits behind master. We just need the one bug fix on this branch -- we'll need another solution.

PR #203, opened 2020-10-06
Fix get merge status

mattcollier:


dlongley (approved):

mattcollier:
23K xblock passes on testcloud.  Merging.

PR #214, opened 2020-11-14
Use fast random id generation.

mattcollier:
taken from https://github.com/ai/nanoid

I noticed we were crossing into C++ land just to generate for some crypto.randomBytes.  Not necessary for this use case.


dlongley (approved):
This seems fine to me.

PR #215, opened 2020-11-15
Simulator fast compare

mattcollier:
I don't think there is any reason this should be using `localCompare` as it does not use any unusual characters.  The profile shows an 10X performance improvement.  Total time before was 3099ms (11% of total), after 318ms (1.16% of total).

**However**, this is likely a breaking change because as this replit shows, the sorted results are not the same.  The two algorithms treat uppercase/lowercase characters differently.  https://repl.it/@mattcollier/GrumpyCompassionateConstant#index.js

See discussion: https://stackoverflow.com/questions/14677060/400x-sorting-speedup-by-switching-a-localecompareb-to-ab-1ab10

## Before
Screenshot from 2020-11-14 20-25-36

## After
Screenshot from 2020-11-14 20-24-49

### DIFF
--- a/lib/consensus.js
+++ b/lib/consensus.js
@@ -1311,7 +1311,7 @@ function _compareSupportSet(set1, set2) {
 }
 
 function _compareEventHashes(a, b) {
-  return a.eventHash.localeCompare(b.eventHash);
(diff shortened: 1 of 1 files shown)


dlongley (approved):
Approving ... the last commit. Note that this PR was targeted at master so we've got everything from the simulator branch too.

PR #216, opened 2020-11-15
Optimize iteration in _tallyBranches.

mattcollier:
This code is in the hot path.  When there are 31 electors (for instance) the `votingEvents` array allocated on L977 is 31 items long and it is the hope that we hit the `some` short circuit sooner rather than later.  This 32 position array is created and destroyed many times.  The new algorithm skips the provisioning of this array, but is still functionally the same.  The full test suite passes with this change.

That said, we never actually fall into the conditional on the new line 988 except in synthetic ledger histories in the test suite. So I would like to discuss the necessity for some of this code existing in the first place.

### DIFF
--- a/lib/consensus.js
+++ b/lib/consensus.js
@@ -974,10 +974,24 @@ function _tallyBranches({
           _computeMostRecentVotes({event, yElectorSet, yByElector});
         }
         // if some voting ancestors haven't computed support yet, defer
-        const votingEvents = Object.values(event._votes);
-        if(votingEvents.some(
-          e => e && e !== event && e._supporting === undefined)) {
-          next.push(event);
+        let pushed = false;
+        for(const v in event._votes) {
+          const e = event._votes[v];
+          // FIXME: remove this conditional, e is undefined in exactly one
+          // place in the test
+          // suite which is some of the findConsensus figures. Specifically
+          // seems to be in relation to a node `b` there.
+          if(!e) {
+            console.log('VVVVVVVVVVVVVv', v);
+            console.log(event._votes);
+          }
+          if(e && e !== event && e._supporting === undefined) {
+            next.push(event);
+            pushed = true;
+            break;
+          }
+        }
+        if(pushed) {
           continue;
         }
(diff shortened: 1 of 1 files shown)


lib/consensus.js
```diff
@@ -974,10 +974,24 @@ function _tallyBranches({
...
-          e => e && e !== event && e._supporting === undefined)) {
-          next.push(event);
+        let pushed = false;
+        for(const v in event._votes) {
+          const e = event._votes[v];
+          // FIXME: remove this conditional, e is undefined in exactly one
+          // place in the test
+          // suite which is some of the findConsensus figures. Specifically
+          // seems to be in relation to a node `b` there.
+          if(!e) {
```
dlongley, line 984: > That said, we never actually fall into the conditional on the new line 988 except in synthetic ledger histories in the test suite. So I would like to discuss the necessity for some of this code existing in the first place.

`e` is false when a byzantine node (fork) is detected, so this is vital. It only happens in "synthetic ledger histories" as you say, or rather, in the test vectors, because it is for the correctness testing for handling byzantine nodes. Our software should never ("can" never) misbehave in this way (fork), so we test this behavior via those test vectors.

lib/consensus.js
```diff
@@ -974,10 +974,24 @@ function _tallyBranches({
           _computeMostRecentVotes({event, yElectorSet, yByElector});
         }
         // if some voting ancestors haven't computed support yet, defer
-        const votingEvents = Object.values(event._votes);
-        if(votingEvents.some(
-          e => e && e !== event && e._supporting === undefined)) {
-          next.push(event);
+        let pushed = false;
```
dlongley, line 977: We use a variable named `defer` in other places in the code for these situations.

```suggestion
        let defer = false;
```

dlongley (changes requested):
Once this is cleaned up, I'll approve, looks good.

gannan08 (approved):

dlongley:
I implemented this, closing this PR.

PR #222, opened 2020-11-24
Fix work scheduling race condition w/new `start` API.

dlongley:


lib/worker/index.js
```diff
@@ -31,14 +31,11 @@ const _ledgerNodeMeta = require('../temporaryLedgerNodeMeta');
 
 const {coolDownPeriod} = config['ledger-consensus-continuity'].gossip;
 
-api.scheduleWork = ({session}) => {
+api.scheduleWork = async ({session}) => {
   // start a consensus session for ledgers
-  const maxTime =
+  const maxAge =
```
mattcollier, line 36: A comment about why the semantics are being changed here would be helpful.  I'm not even sure if maxAge/maxTime is clear, is this really `maxTtl`?

PR #230, opened 2020-12-12
Remove deferred events

dlongley:
@mattcollier -- let me know what you think about some of the changes and FIXMEs here, in particular, whether or not we should be dropping entire batches when only some of the events are problematic. Note that there have been changes to `_validateGraph` ensure that parents referenced by events are validated prior to the events themselves being validated (depth-first validation).

Also, it looks like the `headGeneration` cache/key/functions are no longer used and I've added a commit to remove them. Please let me know what you think.

Note that this PR does not add anything to send the known `blockHeight` during gossip to prevent (well-behaved) peers from sending us events we can't yet process. I'm doing this incrementally to ensure both that the tests keep passing and that we're handling what would be misbehaving peers along the way. Of course, we will eventually need dedicated tests with peers that send us things we can't yet process -- similar to how we need forking peer tests (cc: @gannan08).

### DIFF
--- a/lib/cache/cacheKey.js
+++ b/lib/cache/cacheKey.js
@@ -48,9 +48,6 @@ api.genesis = ledgerNodeId => `g|${_lni(ledgerNodeId)}`;
 api.head = ({creatorId, ledgerNodeId}) =>
   `h|${_lni(ledgerNodeId)}|${_ci(creatorId)}`;
 
-api.headGeneration = ({eventHash, ledgerNodeId}) =>
-  `hg|${_lni(ledgerNodeId)}|${eventHash}`;
-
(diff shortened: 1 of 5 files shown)


mattcollier (approved):
Looks find, I have nothing to add to any of the FIXMEs.  Please try to keep CI green by cleaning up linting errors.

dlongley:
> We will need to audit hidden class transition issues with `<eventRecord>._tmp` ...

Yes, but we should wait on doing anything like that as it may still change/simplify.

> Please try to keep CI green by cleaning up linting errors.

Unfortunately, I'm using those to help me remember things that need fixing. But I'll add FIXMEs instead.

PR #233, opened 2020-12-29
Initialize `externalParents` to an empty array.

gannan08:


dlongley:
I'd rather not generate the extra memory here if we don't need it. Out of potentially thousands of events, only a few may need arrays.

gannan08:
If the assertion is that we rarely create and/or add an `externalParent`, then I'm in agreement that the PR could be closed.

IOW if the overhead for allocating an empty array is greater than the savings from V8 opts, then I'm in agreement that the PR should be closed.

dlongley:
Closing for now because it's likely that this would deoptimize rather than optimize -- and because we don't really know yet -- and because there are conflicts now that aren't worth resolving. It's a minor change later once we go through an optimization phase if we discover this is worth changing.

PR #249, opened 2021-06-15
Add hsigs to notify

aljones15:
1. Adding tests was not in the task so I did not check to ensure the notification was made (tests were done quickly).
2. Addresses issue: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/issues/250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Added
 - `@digitalbazaar/http-client` was added to the project.
 - `lib/client.js` `notifyPeer` API sends an HTTP Signature in each request.
-- `lib/server.js` the notify route now verifies HTTP signatures before creating a notification.
+- `lib/server.js` the notify route now verifies HTTP signatures before accepting a notification.
 
 ### Changed
 - Replaced `axios` with `@digitalbazaar/http-client`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Added
 - `@digitalbazaar/http-client` was added to the project.
 - `lib/client.js` `notifyPeer` API sends an HTTP Signature in each request.
-- `lib/server.js` the notify route now verifies HTTP signatures before issuing a notification.
+- `lib/server.js` the notify route now verifies HTTP signatures before creating a notification.
 
 ### Changed
 - Replaced `axios` with `@digitalbazaar/http-client`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Added
 - `@digitalbazaar/http-client` was added to the project.
+- `lib/client.js` `notifyPeer` API sends an HTTP Signature in each request.
+- `lib/server.js` the notify route now verifies HTTP signatures before issuing a notification.
 
 ### Changed
 - Replaced `axios` with `@digitalbazaar/http-client`
--- a/lib/client.js
+++ b/lib/client.js
@@ -6,6 +6,10 @@
 const _ = require('lodash');
 const {httpsAgent} = require('bedrock-https-agent');
 const {httpClient} = require('@digitalbazaar/http-client');
+const {
+  createAuthzHeader,
+  createSignatureString
+} = require('http-signature-header');
 const https = require('https');
 const bedrock = require('bedrock');
 const {config, util: {BedrockError}} = bedrock;
@@ -124,13 +128,30 @@ exports.notifyPeer = async ({localPeerId, remotePeer}) => {
   const {id: remotePeerId} = remotePeer;
   const url = `${remotePeer.url}/notify`;
   const {'ledger-consensus-continuity': {client: {timeout}}} = config;
+  // the peerId sent to the remote peer is the peerId of the local peer
+  // FIXME: send actual gossip URL instead of `localPeerId` twice
+  const json = {peer: {id: localPeerId, url: localPeerId}};
   // FIXME: use http-signature to authenticate
+  const includeHeaders = [
+    '(created)',
+    '(expires)',
+    '(key-id)',
+    '(request-target)',
+    'digest',
+    'content-type'
+  ];
+  const plainText = createSignatureString({
+    includeHeaders,
+    requestOptions: {url, method: 'POST', keyId}
+  });
+  const headers = {authorization: createAuthzHeader({
+
+  })};
   try {
     await httpClient.post(url, {
       agent: httpsAgent,
-      // the peerId sent to the remote peer is the peerId of the local peer
-      // FIXME: send actual gossip URL instead of `localPeerId` twice
-      json: {peer: {id: localPeerId, url: localPeerId}},
+      headers,
+      json,
       timeout
     });
   } catch(error) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -126,7 +126,7 @@ exports.getHistory = async ({
   return res.data;
 };
 
-exports.notifyPeer = async ({localPeerId, remotePeer}) => {
+exports.notifyPeer = async ({localPeerId, remotePeer, ledgerNodeId}) => {
   const {id: remotePeerId} = remotePeer;
   const url = `${remotePeer.url}/notify`;
   const {'ledger-consensus-continuity': {client: {timeout}}} = config;
@@ -134,7 +134,8 @@ exports.notifyPeer = async ({localPeerId, remotePeer}) => {
   // FIXME: send actual gossip URL instead of `localPeerId` twice
   const json = {peer: {id: localPeerId, url: localPeerId}};
   try {
-    const headers = await _signRequest({keyId: localPeerId, url, json});
+    const headers = await _signRequest(
+      {keyId: localPeerId, url, json, ledgerNodeId});
     await httpClient.post(url, {
       agent: httpsAgent,
       headers,
@@ -275,7 +276,7 @@ function _processHTTPError(error) {
   return {cause, httpStatusCode};
 }
 
-async function _signRequest({keyId, url, json}) {
+async function _signRequest({keyId, url, json, ledgerNodeId}) {
   const created = Math.floor(Date.now() / 1000);
   const expires = created + 5 * 60;
   const includeHeaders = [
@@ -298,7 +299,7 @@ async function _signRequest({keyId, url, json}) {
     requestOptions: {created, expires, url, method: 'POST', headers, keyId}
   });
   const data = new TextEncoder().encode(plainText);
-  const keyPair = await getKeyPair({peerId: keyId});
+  const keyPair = await getKeyPair({ledgerNodeId});
   const sigBuffer = await keyPair.signer().sign({data});
   const signature = Buffer.from(
     sigBuffer, sigBuffer.offset, sigBuffer.length).toString('base64');
--- a/lib/client.js
+++ b/lib/client.js
@@ -130,12 +130,12 @@ exports.notifyPeer = async ({ledgerNodeId, remotePeer}) => {
   const {id: remotePeerId} = remotePeer;
   const url = `${remotePeer.url}/notify`;
   const {'ledger-consensus-continuity': {client: {timeout}}} = config;
+  const key = await getKeyPair({ledgerNodeId});
   // the peerId sent to the remote peer is the peerId of the local peer
   // FIXME: send actual gossip URL instead of `localPeerId` twice
-  const json = {peer: {id: localPeerId, url: localPeerId}};
+  const json = {peer: {id: key.id, url: key.id}};
   try {
-    const headers = await _signRequest(
-      {keyId: localPeerId, url, json, ledgerNodeId});
+    const headers = await _signRequest({url, json, signer: key.signer()});
     await httpClient.post(url, {
       agent: httpsAgent,
       headers,
@@ -146,7 +146,7 @@ exports.notifyPeer = async ({ledgerNodeId, remotePeer}) => {
     const {cause, httpStatusCode} = _processHTTPError(error);
     throw new BedrockError(
       'Could not send peer notification.', 'NetworkError',
-      {localPeerId, httpStatusCode, remotePeerId}, cause);
+      {localPeerId: key.id, httpStatusCode, remotePeerId}, cause);
   }
 };
 
@@ -276,7 +276,7 @@ function _processHTTPError(error) {
   return {cause, httpStatusCode};
 }
 
-async function _signRequest({keyId, url, json, ledgerNodeId}) {
+async function _signRequest({url, json, signer}) {
   const created = Math.floor(Date.now() / 1000);
   const expires = created + 5 * 60;
   const includeHeaders = [
@@ -296,18 +296,24 @@ async function _signRequest({keyId, url, json, ledgerNodeId}) {
   const plainText = createSignatureString({
     includeHeaders,
     headers,
-    requestOptions: {created, expires, url, method: 'POST', headers, keyId}
+    requestOptions: {
+      created,
+      expires,
+      url,
+      method: 'POST',
+      headers,
+      keyId: signer.id
+    }
   });
   const data = new TextEncoder().encode(plainText);
-  const keyPair = await getKeyPair({ledgerNodeId});
-  const sigBuffer = await keyPair.signer().sign({data});
+  const sigBuffer = await signer.sign({data});
   const signature = Buffer.from(
     sigBuffer, sigBuffer.offset, sigBuffer.length).toString('base64');
   const authorization = createAuthzHeader({
     created,
     expires,
     includeHeaders,
-    keyId,
+    keyId: signer.id,
     signature
   });
   return {authorization, ...headers};
--- a/lib/client.js
+++ b/lib/client.js
@@ -126,7 +126,7 @@ exports.getHistory = async ({
   return res.data;
 };
 
-exports.notifyPeer = async ({localPeerId, remotePeer, ledgerNodeId}) => {
+exports.notifyPeer = async ({ledgerNodeId, remotePeer}) => {
   const {id: remotePeerId} = remotePeer;
   const url = `${remotePeer.url}/notify`;
   const {'ledger-consensus-continuity': {client: {timeout}}} = config;
--- a/lib/client.js
+++ b/lib/client.js
@@ -18,6 +18,7 @@ const {pipeline} = require('stream');
 const split2 = require('split2');
 const {validate} = require('bedrock-validation');
 const {URL} = require('url');
+const {getKeyPair} = require('./localPeers');
 
 let EVENTS_VALIDATION_HTTPS_AGENT;
 const EVENTS_VALIDATION_URL_CACHE = new LRU({max: 100, maxAge: 60 * 1000});
@@ -124,17 +125,11 @@ exports.getHistory = async ({
   return res.data;
 };
 
-exports.notifyPeer = async ({localPeerId, remotePeer}) => {
-  const {id: remotePeerId} = remotePeer;
-  const url = `${remotePeer.url}/notify`;
-  const {'ledger-consensus-continuity': {client: {timeout}}} = config;
-  // the peerId sent to the remote peer is the peerId of the local peer
-  // FIXME: send actual gossip URL instead of `localPeerId` twice
-  const json = {peer: {id: localPeerId, url: localPeerId}};
-  // FIXME: use http-signature to authenticate
+const _signRequest = async ({peerId, url, json}) => {
   const includeHeaders = [
     '(created)',
     '(expires)',
+    // (key-id) will include the peerId in the signature itself
     '(key-id)',
     '(request-target)',
     'digest',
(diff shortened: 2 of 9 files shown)


lib/client.js
```diff
@@ -120,17 +126,52 @@ exports.getHistory = async ({
   return res.data;
 };
 
+const _signRequest = async ({keyId, url, json}) => {
```
dlongley, line 129: Switch style and move to bottom of file as a helper.

```suggestion
async function _signRequest({keyId, url, json}) {
```

lib/client.js
```diff
@@ -120,17 +126,52 @@ exports.getHistory = async ({
...
+  const created = Math.floor(Date.now() / 1000);
+  const includeHeaders = [
+    '(created)',
+    // (key-id) will include the peerId in the signature itself
+    '(key-id)',
+    '(request-target)',
```
dlongley, line 135: Should have `(expires)` and `host` as well. Expiration can be 5 minutes into the future.

lib/server.js
```diff
@@ -117,8 +119,29 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+      ];
+      const parsed = parseRequest(
+        {url, method, headers}, {headers: expectedHeaders});
+      // get parsed parameters from from HTTP header and generate signing string
+      const {keyId, signingString, params: {signature: b64Signature}} = parsed;
+      const key = await wrapKeyPair({peerId: keyId});
+      const verifier = key.verifier();
+      const encoder = new TextEncoder();
+      const data = encoder.encode(signingString);
+      const signature = Buffer.from(b64Signature, 'base64');
```
dlongley, line 139: Parsing the signature probably needs to be wrapped in a try/catch.

lib/server.js
```diff
@@ -117,8 +119,29 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+      // verify HTTP signature
+      const verified = await verifier.verify({data, signature});
+      if(!verified) {
+        return res.status(401).end();
```
dlongley, line 143: This needs to be 403. We'd only send a 401 if there was no http signature (vs. it not verifying) -- and if we send a 401, we also have to send a `WWW-Authenticate` header. So, let's just send 403.

lib/signature.js
```diff
@@ -21,6 +21,15 @@ const {
   purposes: {AssertionProofPurpose}
 } = jsigs;
 
+exports.wrapKeyPair = async ({ledgerNodeId, peerId}) => {
```
dlongley, line 24: Let's call this `getKeyPair` instead -- since it isn't wrapping a key pair, it's producing one. Also, this function is called in a number of places in this PR without passing `ledgerNodeId`. This indicates we are probably missing some important tests. Furthermore, `_localPeers.getKeyPair` looks like it takes `ledgerNodeId`, not `peerId`:

https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/master/lib/localPeers.js#L97

And that it only returns the key pair for a local ledger node. So I think there's some kind of conflation going on that needs to get sorted.

dlongley (changes requested):
Thanks! Left some comments. We're close here.

lib/client.js
aljones15: addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/8d345847d6fe7256b303da8a4b1074a2979696d3

lib/server.js
aljones15: I wrapped the entire function in a try/catch and made it a helper here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/2341a04e262986a6d37bfc9c52da3264f595a20a

lib/server.js
aljones15: should be addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/2341a04e262986a6d37bfc9c52da3264f595a20a

lib/signature.js
aljones15: function was renamed, however:

https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/52ac4a1b48e618df0cc649833b1033d41e32a657/lib/client.js#L123-L142

`_client.notifyPeer` only takes in 2 params `{localPeerId, remotePeer}` so it does not immediately at least know the ledgerNodeId (unless I'm missing something). So I used the `localPeerId` to get the key material.

Then in `./lib/localPeers` https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/52ac4a1b48e618df0cc649833b1033d41e32a657/lib/localPeers.js#L267-L278

the private method `_getKeyPair` does take in a `peerId` and as it turns out, ledgerNodeId is not necessary to find the relevant key material. So I added `peerId` as a param of `localPeers.getKeyPair`.

How should I handle this if I'm supposed to not use the `localPeerId`? Find the associated ledger node using: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/52ac4a1b48e618df0cc649833b1033d41e32a657/lib/localPeers.js#L112-L117

perhaps?

lib/server.js
```diff
@@ -117,15 +119,47 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+    const key = await getKeyPair({peerId: keyId});
+    const verifier = key.verifier();
+    const encoder = new TextEncoder();
+    const data = encoder.encode(signingString);
+    const signature = Buffer.from(b64Signature, 'base64');
+    // verify HTTP signature
+    return verifier.verify({data, signature});
+  } catch(e) {
+    logger.error('_verifyRequest errored', {error: e});
+    return false;
```
dlongley, line 159: I think we want to surface errors that prevented verification from even running (vs. just a bad verify) to the client, e.g., if their signature is in the wrong format.

The previous code would have returned a 500 if their signature format was bad, which we don't want (implying the server had an issue, which it didn't). But this new code will hide too much. What we want is, if the client data is bad/the mistake is on them, we want to return an appropriate error code and error message.

dlongley (changes requested):

lib/signature.js
dlongley: So we should modify `sendNotification` so it requires the `ledgerNodeId`. Only a `Worker` instance calls it, so it can pass it through.

aljones15: addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/32905e6cd329fbbdf09a4181178cad561d9c1c5a
and https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/7c1fd0d1925a8caae15c8a0e7e2ab0ec52de4c4e

lib/server.js
aljones15: first attempt here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/9bad307ee998b689b905f616b4a3c18eefb88db2

lib/client.js
```diff
@@ -126,15 +126,16 @@ exports.getHistory = async ({
   return res.data;
 };
 
-exports.notifyPeer = async ({localPeerId, remotePeer}) => {
+exports.notifyPeer = async ({localPeerId, remotePeer, ledgerNodeId}) => {
```
dlongley, line 129: It seems to me that all of this must be driven off of the `ledgerNodeId`, so we shouldn't pass the `localPeerId` here. It might even be in conflict with whatever is retrieved via `_getKeyPair`. It looks like `_getKeyPair` returns the `localPeerId` anyway (and it's guaranteed to match): https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/blob/7c1fd0d1925a8caae15c8a0e7e2ab0ec52de4c4e/lib/localPeers.js#L271

```suggestion
exports.notifyPeer = async ({ledgerNodeId, remotePeer}) => {
```

lib/client.js
```diff
@@ -266,3 +275,40 @@ function _processHTTPError(error) {
   }
   return {cause, httpStatusCode};
 }
+
+async function _signRequest({keyId, url, json, ledgerNodeId}) {
```
dlongley, line 279: So I think this is a bit tangled to have both the `ledgerNodeId` and the `keyId` here. I think we should pass in the signer API here instead of the `ledgerNodeId` and fetch it externally.

lib/server.js
```diff
@@ -117,15 +120,57 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+    const signature = Buffer.from(b64Signature, 'base64');
+    // verify HTTP signature
+    return verifier.verify({data, signature});
+  } catch(e) {
+    throw new BedrockError(
+      'Signature Verification failed',
+      'DataError',
+      {public: true, httpStatusCode: 403},
+      e
+    );
```
dlongley, line 170:
```suggestion
    throw new BedrockError(
      'Signature verification failed.', 'DataError',
      {public: true, httpStatusCode: 403}, e);
```

lib/worker/gossip.js
```diff
@@ -86,12 +86,16 @@ exports.runGossipCycle = async ({worker, needsGossip}) => {
   return {mergeEventsReceived, busy};
 };
 
-exports.sendNotification = async ({localPeerId, peerSelector} = {}) => {
+exports.sendNotification = async ({
+  localPeerId,
+  peerSelector,
+  ledgerNodeId
+} = {}) => {
```
dlongley, line 93:
```suggestion
exports.sendNotification = async ({ledgerNodeId,  peerSelector} = {}) => {
```

lib/worker/gossip.js
```diff
@@ -86,12 +86,16 @@ exports.runGossipCycle = async ({worker, needsGossip}) => {
...
   // notify all selected peers in parallel
   const peers = await peerSelector.selectNotifyPeers();
   const promises = [];
   for(const peer of peers) {
-    promises.push(_notifyPeer({localPeerId, peer}));
+    promises.push(_notifyPeer({localPeerId, peer, ledgerNodeId}));
```
dlongley, line 98:
```suggestion
    promises.push(_notifyPeer({ledgerNodeId, peer}));
```

lib/worker/gossip.js
```diff
@@ -274,11 +278,11 @@ async function _diff({worker, eventHashes}) {
   return result;
 }
 
-async function _notifyPeer({localPeerId, peer}) {
+async function _notifyPeer({localPeerId, peer, ledgerNodeId}) {
```
dlongley, line 281:
```suggestion
async function _notifyPeer({ledgerNodeId, peer}) {
```

lib/worker/gossip.js
```diff
@@ -274,11 +278,11 @@ async function _diff({worker, eventHashes}) {
...
   // FIXME: fix access of private `_peer`
   const remotePeer = {id: peer._peer.id, url: peer._peer.url};
   try {
-    await _client.notifyPeer({localPeerId, remotePeer});
+    await _client.notifyPeer({localPeerId, remotePeer, ledgerNodeId});
```
dlongley, line 285:
```suggestion
    await _client.notifyPeer({ledgerNodeId, remotePeer});
```

dlongley (changes requested):

lib/client.js
aljones15: addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/e90e694686bc49f31c52bbdfa63fe6a3d3dc6f37

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Added
 - `@digitalbazaar/http-client` was added to the project.
+- `lib/client.js` `notifyPeer` API sends an HTTP Signature in each request.
+- `lib/server.js` the notify route now verifies HTTP signatures before creating a notification.
```
dlongley, line 12:
```suggestion
- `lib/server.js` the notify route now verifies HTTP signatures before accepting a notification.
```

lib/server.js
```diff
@@ -117,15 +120,54 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
     routes.notify, brRest.when.prefers.ld,
     validate('continuity-server.notification'),
     asyncHandler(async (req, res) => {
-      // FIXME: we need to authenticate `peerId`; http-signatures should be
-      // added here
+      const verified = await _verifyRequest({req});
+      if(!verified) {
```
dlongley, line 124: So in addition to verifying the request, we also need to check whether the `keyId` in the signature matches the expected `peer.id` in the body of the request. Otherwise, we should also send 403. So we'll want to do something like this:

```suggestion
      const {verified, keyId} = await _verifyRequest({req});
      if(!(verified && req.body.peer.id === keyId)) {
```

And make `_verifyRequest` return back more than just the verified status, but also which `keyId` signed.

lib/worker/Worker.js
```diff
@@ -204,8 +204,8 @@ module.exports = class Worker {
   }
 
   async notifyPeers() {
-    const {localPeerId, peerSelector} = this;
-    return sendNotification({localPeerId, peerSelector});
+    const {localPeerId, peerSelector, ledgerNodeId} = this;
+    return sendNotification({localPeerId, peerSelector, ledgerNodeId});
```
dlongley, line 208:
```suggestion
    const {ledgerNodeId, peerSelector} = this;
    return sendNotification({ledgerNodeId, peerSelector});
```

test/mocha/050-client-api.js
```diff
@@ -3,20 +3,42 @@
...
 'use strict';
 
-const {_client} = require('bedrock-ledger-consensus-continuity');
+const {_client, _localPeers} = require('bedrock-ledger-consensus-continuity');
 const {config, util: {BedrockError}} = require('bedrock');
 
 describe('Client API', () => {
   describe('notifyPeer', () => {
+    // NOTE: this name is really bad. peerId is actually a database record
+    // with tons of information including private key material
```
dlongley, line 12: `peerId` is a database record and not a peer ID? That doesn't sound right. Can you look at this?

test/mocha/050-client-api.js
```diff
@@ -3,20 +3,42 @@
...
+    let peerId = null;
+    const ledgerNodeId = 'foo';
+    before(async () => {
+      ({peerId} = await _localPeers.generate({ledgerNodeId}));
+    });
+    it('throws a NotFoundError if peerId is not found', async () => {
```
dlongley, line 18: This is the case where `ledgerNodeId` is not found, right?

test/mocha/050-client-api.js
```diff
@@ -3,20 +3,42 @@
...
-      const localPeerId = 'foo';
+      const localPeerId = peerId.localPeer.peerId;
       const remotePeer = {
         id: 'https://127.0.0.1',
         url: 'https://127.0.0.1'
       };
       let err;
       try {
-        await _client.notifyPeer({localPeerId, remotePeer});
+        await _client.notifyPeer({localPeerId, remotePeer, ledgerNodeId});
```
dlongley, line 41:
```suggestion
        await _client.notifyPeer({ledgerNodeId, remotePeer});
```

test/mocha/050-client-api.js
```diff
@@ -3,20 +3,42 @@
...
+      } catch(e) {
+        err = e;
+      }
+      should.exist(err);
+      err.name.should.equal('NotFoundError');
+    });
+
     it('throws a NetworkError on connection refused', async () => {
-      const localPeerId = 'foo';
+      const localPeerId = peerId.localPeer.peerId;
```
dlongley, line 34: I don't think this is used anymore, right?

```suggestion
```

test/mocha/050-client-api.js
```diff
@@ -32,6 +54,24 @@ describe('Client API', () => {
       err.cause.details.should.have.property('port');
       err.cause.details.code.should.equal('ECONNREFUSED');
     });
+
+    it('should notify peer', async () => {
+      const localPeerId = peerId.localPeer.peerId;
```
dlongley, line 59:
```suggestion
```

test/mocha/050-client-api.js
```diff
@@ -32,6 +54,24 @@ describe('Client API', () => {
...
+      const remotePeer = {
+        id: localPeerId,
+        url: localPeerId
+      };
+      let err;
+      try {
+        // notifyPeer does not return anything
+        await _client.notifyPeer({localPeerId, remotePeer, ledgerNodeId});
```
dlongley, line 67:
```suggestion
        await _client.notifyPeer({ledgerNodeId, remotePeer});
```

dlongley (changes requested):
Ok, nearly there, I think.

test/mocha/050-client-api.js
aljones15: yes and addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/0c2ae776d0a3bce31ea8530262f5aec4fd5c79e2

test/mocha/050-client-api.js
aljones15: Looked into it and made an issue: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/issues/250

lib/localPeers.js
```diff
@@ -106,7 +105,6 @@ api.getKeyPair = async ({ledgerNodeId}) => {
     _pruneCacheTimer = setTimeout(
       _pruneCache, LOCAL_PEER_KEY_PAIR_CACHE_MAX_AGE);
   }
-
```
dlongley, line 108: Let's keep this whitespace.

test/mocha/050-client-api.js
dlongley: Let's fix that issue here now (seems like a one line fix) and remove this note.

test/mocha/050-client-api.js
```diff
@@ -32,6 +53,24 @@ describe('Client API', () => {
...
+      };
+      let err;
+      try {
+        // notifyPeer does not return anything
+        await _client.notifyPeer({ledgerNodeId, remotePeer});
+      } catch(e) {
+        err = e;
+      }
+      should.not.exist(err);
+      // FIXME look for the actual notification here
```
dlongley, line 71: Can we address this FIXME or remove it and add an issue to improve the assertions on notification tests?

lib/client.js
```diff
@@ -266,3 +275,46 @@ function _processHTTPError(error) {
   }
   return {cause, httpStatusCode};
 }
+
+async function _signRequest({url, json, signer}) {
```
dlongley, line 279: Thinking about this more, we should move `_signRequest` and `_verifyRequest` into the signature.js helper -- and make them use a common array for the headers so that they don't get de-synced. It also seems like the right place to put the crypto details vs. having them spread across client/server.

dlongley (changes requested):

test/mocha/050-client-api.js
aljones15: addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/ac9aec6e68f0b3681c6b1508bd2ba7e40150f39e and the commit after it.

lib/localPeers.js
aljones15: addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/249/commits/b18261564956b343ca5b00925f3d5c551aa482b1

lib/server.js
```diff
@@ -117,15 +120,55 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
...
+      'digest',
+      'content-type',
+      'host'
+    ];
+    const parsed = parseRequest(
+      {url, method, headers}, {headers: expectedHeaders});
+    // get parsed parameters from from HTTP header and generate signing string
+    const {keyId, signingString, params: {signature: b64Signature}} = parsed;
+    // returns the key bytes
+    const key = new Ed25519VerificationKey2018({
```
dlongley, line 152: Note: There's a `_getPublicKey` helper in signature.js (once this is moved) that may help here.

lib/signature.js
```diff
@@ -46,7 +56,68 @@ exports.sign = async ({event, ledgerNodeId}) => {
     suite: new MergeEventEd25519Signature2018({key}),
     purpose: new AssertionProofPurpose()
   });
-  return {signed, peerId: localPeer.peerId};
+  return {signed, peerId: key.id};
+};
+
+const coveredContent = [
```
dlongley, line 62:
```suggestion
const COVERED_CONTENT = [
```

lib/signature.js
```diff
@@ -46,7 +56,68 @@ exports.sign = async ({event, ledgerNodeId}) => {
...
+ * @param {Array<string>} [options.includeHeaders = coveredContent] - Which
+ *   headers to include in the signature.
+ *
+ * @returns {Promise<object>} Returns the headers for the request.
+ */
+exports.signRequest = async ({
+  url,
+  json,
+  signer,
+  includeHeaders = coveredContent
```
dlongley, line 89:
```suggestion
  includeHeaders = COVERED_CONTENT
```

lib/signature.js
```diff
@@ -46,7 +56,68 @@ exports.sign = async ({event, ledgerNodeId}) => {
...
+];
+
+/**
+ * Signs a request with a valid HTTP Signature header.
+ *
+ * @param {object} options - Options to use.
+ * @param {URL} options.url - The request URL.
+ * @param {object} options.json - The request body.
+ * @param {object} options.signer - A signer for the signature.
+ * @param {Array<string>} [options.includeHeaders = coveredContent] - Which
```
dlongley, line 80:
```suggestion
 * @param {Array<string>} [options.includeHeaders = COVERED_CONTENT] - Which
```

lib/signature.js
```diff
@@ -93,6 +164,44 @@ exports.verify = async ({event}) => {
   return {controller: publicKeyId};
 };
 
+/**
+ * Verifies a request's HTTP signature header.
+ *
+ * @param {object} options to use.
+ * @param {object} options.req - An express Request object.
+ * @param {Array<string>} [options.expectedHeaders = coveredContent] - Which
```
dlongley, line 172:
```suggestion
 * @param {Array<string>} [options.expectedHeaders = COVERED_CONTENT] - Which
```

lib/signature.js
```diff
@@ -93,6 +164,44 @@ exports.verify = async ({event}) => {
...
+ *   headers are expected to be in the request.
+ *
+ * @returns {Promise<object>} The result of verification and the keyId.
+ */
+exports.verifyRequest = async ({req, expectedHeaders = coveredContent}) => {
```
dlongley, line 177:
```suggestion
exports.verifyRequest = async ({req, expectedHeaders = COVERED_CONTENT}) => {
```

dlongley (approved):
Thanks!

dlongley:
Closes #250.

PR #253, opened 2021-06-19
Upgrade to use 2020 cryptosuites.

msporny:


lib/signature.js
```diff
@@ -138,15 +139,26 @@ exports.verify = async ({event}) => {
...
+
+  ///////////////////// FIXME: The signer shouldn't have to be specified
+  key.signer = {
+    sign: () => {
+      console.log(
+        'ERROR: MergeEventEd25519Signature2020.signer called during verify().');
+      return 'INVALID';
+    }
+  };
+  /////////////////////
```
msporny, line 152: If this code isn't here, the code throws creating the suite for verification. The Ed25519Signature2020 tries to create a signer even when a private key isn't supplied. What's up with that? /cc @dmitrizagidulin

dlongley: +1 agree that it shouldn't be needed.

dlongley (approved):
LGTM, other than the comment about removing `signer` stuff that shouldn't be necessary. Thanks!

PR #254, opened 2021-06-20
Only ignore duplicate event error when inserting genesis event.

dlongley:
I was looking around for the source of an off-by-one error and this came up as a possible problem since worker state is modified (`nextLocalEventNumber` is incremented) for an event that may then be considered a duplicate. However, this doesn't actually ever happen (AFAICT) in practice when adding an event via a worker (the same function is used to add the genesis event without a worker), and if you force it to happen, it happily carries on without a problem. So this fix did not the solve the problem I was looking into.

Regardless, I think we should only be ignoring the duplicate when it is for the genesis event (a failed ledger node creation attempt occurred and we can try again later). I don't see any reason for a duplicate error to occur on an up and running node and we should throw if this happens instead of allowing potentially bad state (even if it doesn't seem fatal) to get into the system.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -595,7 +595,6 @@ async function _writeEvent({event, eventHash, genesis, ledgerNode, worker}) {
       'must not be added outside of a work session.');
   }
 
-  const ledgerNodeId = ledgerNode.id;
   const {storage} = ledgerNode;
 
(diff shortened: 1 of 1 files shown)


dlongley:
This PR also resolves an old FIXME by removing what amounts to dead code.

gannan08 (approved):

PR #257, opened 2021-06-24
MongoDB Operation Queue

gannan08:


lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,267 @@
...
+      this.timeoutId = null;
+    }
+    this.running = false;
+    this.stopped = null;
+    this._resolveStopped();
+  }
+
+  /**
+   * Adds an operation to be written. If this OperationWriter is running and the
+   * operation causes the queue to reach the configured capacity threshold, then
```
dlongley, line 97: Should this be:

```suggestion
   * operation causes the queue to reach the configured `highWaterMark`, then
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,267 @@
...
+   * it will automatically flush operations to the underlying database via a
+   * bulk insert.
+   *
+   * @param {object} options - Options to use.
+   * @param {object} options.operation - The operation to write.
+   * @param {object} options.meta - The metadata to write.
+   *
```
dlongley, line 104:
```suggestion
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,267 @@
...
+    } catch(e) {
+      if(!database.isDuplicateError(e)) {
+        throw e;
+      }
+    }
+  }
+
+  /**
+   * Caches the length of the database in queue in memory.
+   *
```
dlongley, line 210:
```suggestion
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,267 @@
...
+   */
+  async _cacheDatabaseQueueLength() {
+    const fn = async () => {
+      const {ledgerNodeId} = this;
+      const projection = {_id: 1};
+      const count = await this.collection.find({
+        ledgerNodeId
+      }, {projection}).count();
```
dlongley, line 218:
```suggestion
      const count = await this.collection.count({ledgerNodeId});
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,267 @@
...
+      this.databaseOperationQueueLength = count;
+    };
+
+    try {
+      await PROMISE_QUEUE.memoize({
+        key: 'updateCount',
+        fn
+      });
```
dlongley, line 226:
```suggestion
      await PROMISE_QUEUE.memoize({key: 'updateCount', fn});
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,267 @@
...
+   * Checks if the operation is a duplicate.
+   *
+   * @param {object} operation - A webledger operation record.
+   * @param {object} operation.meta - The metadata of an operation.
+   */
+  _isDuplicateOperation({meta}) {
+    const {operationHash} = meta;
+    if(!operationHash) {
+      throw new Error('`meta` must contain "operationHash".');
+    }
```
dlongley, line 257: Why not just pass `operationHash`? And why do we have to check `meta` for it?

dlongley (commented):
Just some quick comments for now. I haven't had time to do a thorough review yet.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
+/*!
+ * Copyright (c) 2021 Digital Bazaar, Inc. All rights reserved.
+ */
+const database = require('bedrock-mongodb');
+const logger = require('../logger');
+const LRU = require('lru-cache');
+const {config, util: {BedrockError}} = require('bedrock');
+const {LruCache} = require('@digitalbazaar/lru-memoize');
+
+const PROMISE_QUEUE = new LruCache({max: 10, disposeOnSettle: true});
```
dlongley, line 10: What's the rationale for the max size here? Will the same LRU cache be servicing more than 10 concurrent requests at a time? Should the name of the promise queue be more targeted for which function it is for? Or is it intended to cover multiple different functions? It looks like it is only used in one place right now, so I recommend making its name more specific.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+
+// Error Messages
+const DATABASE_QUEUE_LENGTH_ERROR =
+  'Non-critical error: failed to cache length of operation queue on disk.';
+
+class OperationWriter {
+  constructor({ledgerNodeId, highWaterMark = 100, timeout = 1000} = {}) {
```
dlongley, line 17: Please add a comment explaining the rationale for these defaults.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+      return;
+    }
+    // cancel any scheduled flush
+    if(timeoutId !== null) {
+      clearTimeout(timeoutId);
+      this.timeoutId = null;
+    }
+    this.running = false;
+    this.stopped = null;
+    this._resolveStopped();
```
dlongley, line 92: I'm just skimming over most of the stop mechanics here and presuming it came from another place that is reasonably well tested or what we have here has some decent tests or something. It's probably not so critical that if it fails it will create irreparable harm to a peer's membership in a ledger.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+  /**
+   * Adds an operation to be written. If this OperationWriter is running and the
+   * operation causes the queue to reach the configured `highWaterMark`, then
+   * it will automatically flush operations to the underlying database via a
+   * bulk insert.
+   *
+   * @param {object} options - Options to use.
+   * @param {object} options.operation - The operation to write.
+   * @param {object} options.meta - The metadata to write.
+   */
```
dlongley, line 104: `[options.forceFlush=false]` isn't documented.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+  async add({operation, meta, forceFlush = false} = {}) {
+    const {queue, running, highWaterMark, MAX_OPERATION_QUEUE_LENGTH} = this;
+
+    if(this.databaseOperationQueueLength >= MAX_OPERATION_QUEUE_LENGTH) {
+      // we intentionally do not await the promise, the length can be updated
+      // after the error is returned
```
dlongley, line 110:
```suggestion
      // after the error is returned without ill effects
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+      this._cacheDatabaseQueueLength().catch(e => {
+        logger.debug(DATABASE_QUEUE_LENGTH_ERROR, {error: e});
+      });
+      throw new BedrockError(
+        'The node is not accepting operations. Try again later.',
```
dlongley, line 115: Not sure if we want to say `peer` here instead. Can't remember, your call.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+        'OperationError', {
+          httpStatusCode: 503,
+          public: true,
+        });
+    }
+
+    if(this._isDuplicateOperation(meta)) {
```
dlongley, line 122: It would be good to mention that this check is a soft duplicate check and that other processes could have accepted the same operation -- and, if appropriate, mention that the same process could accept the same operation if another one was pushed out of the LRU cache. Perhaps the explanation will be that the max queue length needs to be less than the LRU cache size for this to work properly -- but I'll note that the max queue length seems to be a soft check as well, so there seems like there could be bugs at the edges.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+    queue.push({operation, meta});
+
+    // do not flush if not writer is not running
+    if(!running) {
+      return;
+    }
+
+    // force flush if required
+    if(forceFlush) {
+      await this.flush();
```
dlongley, line 144: Should this return without bothering to schedule a flush since one just happened, or not? If not, can you leave a comment as to why not? 

```suggestion
      await this.flush();
      // return as flush was forced; no need to potentially schedule a flush
      return;
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    // if the queue length matches the highWaterMark, schedule an immediate
+    // flush; only schedule this when they match do prevent repeatedly
```
dlongley, line 148:
```suggestion
    // flush; only schedule this when they match to prevent repeatedly
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+      this.timeoutId = null;
+    }
+
+    // save current queue and replace it with a new one
+    const queue = this.queue;
+    this.queue = [];
+
+    if(queue.length === 0) {
+      return;
+    }
```
dlongley, line 181: Should checking the queue length happen first so as to avoid instantiating a new array unnecessarily?

```suggestion
    if(queue.length === 0) {
      return;
    }

    // save current queue and replace it with a new one
    const queue = this.queue;
    this.queue = [];
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+    }));
+
+    try {
+      const writeOptions = {...database.writeOptions, ordered: false};
+      await this.collection.insertMany(records, writeOptions);
+      await this._cacheDatabaseQueueLength().catch(e => {
+        logger.debug(DATABASE_QUEUE_LENGTH_ERROR, {error: e});
+      });
+    } catch(e) {
+      if(!database.isDuplicateError(e)) {
```
dlongley, line 201:
```suggestion
      // ignore duplicates; other processes that do not share memory
      // with this one could have concurrently written duplicate ops;
      // these duplicate ops must be filtered out later by the worker
      // that puts ops into events
      if(!database.isDuplicateError(e)) {
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+   * @param {object} options.operation - The operation to write.
+   * @param {object} options.meta - The metadata to write.
+   */
+  async add({operation, meta, forceFlush = false} = {}) {
+    const {queue, running, highWaterMark, MAX_OPERATION_QUEUE_LENGTH} = this;
+
+    if(this.databaseOperationQueueLength >= MAX_OPERATION_QUEUE_LENGTH) {
+      // we intentionally do not await the promise, the length can be updated
+      // after the error is returned
+      this._cacheDatabaseQueueLength().catch(e => {
```
dlongley, line 111: While this is doing caching, I think that's an internal detail of the function. What it's actually doing is updating the queue length, right? So I think it should be named that. I had to read all the code here to figure out why it was caching anything and went looking to see if the cache was ever read and got confused. I know understand that multiple calls may be made to update the queue length concurrently -- and you're trying to effectively debounce that. So, I think if you name this `_updateDatabaseQueueLength` it will be clearer what the point of the function is.

```suggestion
      this._updateDatabaseQueueLength().catch(e => {
```

dlongley: It would also be good to add a note here as to why the length gets updated in this case. I had to reason through why this would be called here. I saw that the only other place that the queue length gets updated is the intuitive spot where the ops get flushed to the database. It looks like we have no feedback from when the queue gets drained, so this is a place where we test to see if anything has drained because we're over the limit.

I'll note that if the queue has drained we'll still reject the current op because we throw without waiting for the fresh queue length. We should mention that we know that and it's intentional here (unless it wasn't) -- for simplicity of design. We should also mention that since we don't get any feedback on when the queue is draining, we make the update call to get a fresh value for the database queue length whenever we're at/over the limit.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+      if(!database.isDuplicateError(e)) {
+        throw e;
+      }
+    }
+  }
+
+  /**
+   * Caches the length of the database in queue in memory.
+   */
+  async _cacheDatabaseQueueLength() {
```
dlongley, line 210:
```suggestion
  async _updateDatabaseQueueLength() {
```

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,255 @@
...
+      });
+    } catch(e) {
+      if(!database.isDuplicateError(e)) {
+        throw e;
+      }
+    }
+  }
+
+  /**
+   * Caches the length of the database in queue in memory.
```
dlongley, line 208:
```suggestion
   * Updates the length of the database in queue in memory if it hasn't been
   * updated too recently. Updates to the database queue length are debounced
   * for better performance. Getting a consistent value for the database queue
   * length is also not part of the design, as multiple isolated processes may be
   * writing to the database queue concurrently and another isolated process
   * may be draining it concurrently. Therefore, design here presumes there
   * will be not strong consistency when tracking the database length; rather,
   * it allows for some operations to be refused to be queued when under load,
   * even if it may have been technically possible for those operations to be
   * queued with a strongly consistent design. This is considered to be an
   * acceptable condition given the performance benefits from not requiring
   * strong consistency and with the observation that false positives will be
   * in the margins on a taxed system.
```

dlongley (changes requested):
Reviewed the OperationWriter -- will do more later on the OperationQueue.

lib/cache/operations.js
```diff
@@ -3,11 +3,28 @@
  */
```
dlongley, line 3: I think we should move this file out of the `cache` dir and into something like `operationWriter.js` at the top level. We want to keep removing things from the cache dir until it's empty and we can remove it entirely.

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+
+const _continuityConstants = require('../continuityConstants');
+const _util = require('../util');
+const database = require('bedrock-mongodb');
+const logger = require('../logger');
+const pLimit = require('p-limit');
+
+// Controls the max concurrency for a given number of promises executing. This
+// is done to ensure we reduce pressure on the NodeJS Promise Queue and limit
+// the number concurrent I/O operations. The number 25 was arbitrarily picked.
```
dlongley, line 14: Let's do better than arbitrary. Is `25` truly out of a hat? This is used when validating operations (which we should mention). I see that we may have to validate `250` concurrently at any time, which is the default max cached queue size. Perhaps we could talk about the number `25` being a 1/10th the max number of concurrent ops being validated?   If we're going to be choosing semi-arbitrary things, I'd like us to be using powers of ten in some way because they are easier to reason about and we can at least point to that.

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+// Controls the maximum number of documents to pull from the database and cache
+// locally. The Math.max(DEFAULT_MAX_CACHED_QUEUE_SIZE, MAX_OPS_PER_REG_EVENT)
+// will be the final value used to limit the number of documents pulled from
+// the database.
+const DEFAULT_MAX_CACHED_QUEUE_SIZE = 250;
+
+// Error Messages
+const OPERATIONS_REMOVAL_ERROR = 'Failed to remove operations from queue.';
+const OPERATION_VALIDATION_ERROR = 'Non-critical operation validation error. ' +
+  'Invalid head of operation queue.';
```
dlongley, line 25:
```suggestion
const OPERATION_VALIDATION_ERROR =
  'Non-critical operation validation error. Invalid head of operation queue.';
```

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+
+module.exports = class OperationQueue {
+  constructor({worker}) {
+    // continuity worker
```
dlongley, line 29: I'd rather us talk about it being the consensus worker -- or the session worker than talking about continuity as that's the name of just the low-level consensus algorithm itself. The worker runs the pipeline, does gossip, creates merge events and blocks ... all that.

```suggestion
    // session worker
```

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+    // constants
+    this.MAX_OPERATIONS = _continuityConstants.events.maxOperations;
+    this.MAX_CACHED_QUEUE_SIZE = Math.max(
+      DEFAULT_MAX_CACHED_QUEUE_SIZE,
+      this.MAX_OPERATIONS);
+  }
+
+  /**
+   * Returns whether or not there is another chunk of operations that can
+   * be put into a local regular event. This MUST NOT be called concurrently.
```
dlongley, line 65: `or not` is superfluous.

```suggestion
   * Returns whether there is another chunk of operations that can be put
   * into a local regular event. This MUST NOT be called concurrently.
```

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+      await this.removing.catch(e => {
+        logger.error(OPERATIONS_REMOVAL_ERROR, {ledgerNodeId, error: e});
+      });
+    }
+
+    // if the queue is not initialized or the head of the queue is at the end
+    // of the queue, then no next chunk of operations cached yet, so create one
+    // that will be cached when `getNextChunk` is called and its operations are
+    // retrieved
+    if(!(Array.isArray(this.queue) && (this.head + 1) < this.queue.length)) {
```
dlongley, line 88: I think this is clearer for checking whether "head is before the end of the queue".

```suggestion
    const end = this.queue.length - 1;
    if(!(Array.isArray(this.queue) && this.head < end)) {
```

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+    if(this.removing) {
+      await this.removing.catch(e => {
+        logger.error(OPERATIONS_REMOVAL_ERROR, {ledgerNodeId, error: e});
+      });
+    }
+
+    // if the queue is not initialized or the head of the queue is at the end
+    // of the queue, then no next chunk of operations cached yet, so create one
+    // that will be cached when `getNextChunk` is called and its operations are
+    // retrieved
```
dlongley, line 87:
```suggestion
    // if the queue is not initialized with the head of the queue before the end
    // of the queue, then no next chunk of operations is cached yet, so create one
    // that will be cached when `getNextChunk` is called and its operations are
    // retrieved
```

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+    if(!(Array.isArray(this.queue) && (this.head + 1) < this.queue.length)) {
+      this.head = 0;
+      this.queue = null;
+
+      const {ledgerNodeId} = this;
```
dlongley, line 92: Already did this above.

```suggestion
```

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+      const projection = {_id: 0, meta: 1, operation: 1};
+
+      if(this.fetchRecords === null) {
+        const {MAX_CACHED_QUEUE_SIZE} = this;
+        this.fetchRecords = this.collection.find({ledgerNodeId}, {projection})
```
dlongley, line 97: Only need `projection` within the conditional.

```suggestion
      if(this.fetchRecords === null) {
        const {MAX_CACHED_QUEUE_SIZE} = this;
        const projection = {_id: 0, meta: 1, operation: 1};
        this.fetchRecords = this.collection.find({ledgerNodeId}, {projection})
```

lib/worker/OperationQueue.js
dlongley: Also, this feels a little OBO prone since we're not just advancing the head to the length of the queue, but I'm going to ignore that. To be clear, it seems odd to me to say that when the head is pointing to the last element in the queue we will reset it to `0` ... what about that last element?

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+
+    const {head, opCount} = await this._getHead();
+
+    if(head === null) {
+      this.fetchRecords = null;
+      logger.debug('Unable to find a valid head of operation queue.', {
+        ledgerNodeId
+      });
+      return false;
+    }
```
dlongley, line 128: This doesn't quite feel clean -- like we've got some state that is partitioned that doesn't need to be. We initialized `head` back to `0` and `queue` back to `null` but we potentially kept this `fetchRecords` thing around as internal state from some past thing... and awaited it ... and now we're calling `this._getHead()` on that. Can `fetchRecords` just live in here locally? Why does it need to live beyond this function call as instance state?

Sorry, having a hard time following this bit and it seems like it should be simpler -- I may be missing something.

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+      const bulkOp = this.collection.initializeUnorderedBulkOp();
+
+      for(const {meta} of this.removedOperations) {
+        bulkOp.find({
+          ledgerNodeId,
+          'meta.operationHash': meta.operationHash
+        }).delete();
+      }
+
+      await bulkOp.execute();
```
dlongley, line 319: I'm unfamiliar with this API -- going to trust you here.

lib/worker/OperationQueue.js
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  async _validateOperation({operation, basisBlockHeight}) {
+    const {meta: {basisBlockHeight: operationBbh}} = operation;
+    const {ledgerNode} = this;
+
+    // Operations are validated at the latest block height of the ledger node
+    // before being accepted into the queue. We can return early when the ledger
+    // node has not advanced its block height, therefore preventing
+    // duplicating the execution of the same validation check.
```
dlongley, line 350: When using full sentences/paragraphs, etc. use multiline comment syntax.

```suggestion
    /* Operations are validated at the latest block height of the ledger node
    before being accepted into the queue. We can return early when the ledger
    node has not advanced its block height, therefore preventing
    duplicating the execution of the same validation check. */
```

dlongley (changes requested):
Ok, reviewed through OperationQueue but got a little stuck. Hopefully my comments are helpful and there's some quick refactoring that can be done to simplify an then I'll review again. We're getting really close here.

lib/operations.js
```diff
@@ -22,7 +22,7 @@ module.exports = api;
  *
  * @return a Promise that resolves once the operation completes.
  */
-api.add = async ({meta, operation, ledgerNode}) => {
+api.add = async ({meta, operation, forceFlush, ledgerNode}) => {
```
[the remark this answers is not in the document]
gannan08: If this is the case, then an issue should be created to make this change both here and `br-ledger-node`. I'm punting on this for the PR, this is an internal bedrock-ledger* change.

lib/worker/OperationQueue.js
gannan08: This is done to prevent multiple database calls when the function is called back-to-back in rapid succession.

lib/worker/OperationQueue.js
gannan08: See: https://docs.mongodb.com/manual/reference/method/db.collection.initializeUnorderedBulkOp/

lib/events.js
```diff
@@ -182,22 +181,20 @@ api.add = async (
...
   // get chunk of operations to put into the event
-  // FIXME: `basisBlockHeight` should be determined by the worker
-  const {basisBlockHeight, hasMore, operations} = await queue.getNextChunk();
+  const {hasMore, operations} = await queue.getNextChunk();
 
   const {head} = worker;
   const baseEvent = {
     '@context': config.constants.WEB_LEDGER_CONTEXT_V1_URL,
     type: 'WebLedgerOperationEvent',
-    basisBlockHeight,
```
gannan08, line 200: This wasn't raised, but this is my biggest concern. The previous implementation depended on the queue to receive the proper basis block height. The new behavior assumes that `worker.consensusState.blockHeight` is the same block height that the operations received were validated at. What guarantees that in the worker? Is it the case that the `worker.consensusState.blockHeight` will always match the blockheight the operations were validated at?

lib/operations.js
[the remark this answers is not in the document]
gannan08: Seems the change has been made already. Updated to match recommendation.

lib/events.js
dlongley: I thought the queue was setup to run operation validation at a certain block height -- and this block height was provided by the worker. If that's true -- and we don't use left over or "stale" operation validations from a previous block height, we should be good. We should discuss more at some point.

lib/cache/OperationWriter.js
```diff
@@ -7,13 +7,33 @@ const LRU = require('lru-cache');
 const {config, util: {BedrockError}} = require('bedrock');
 const {LruCache} = require('@digitalbazaar/lru-memoize');
 
-const PROMISE_QUEUE = new LruCache({max: 10, disposeOnSettle: true});
+/* Note: Max size is 1. This is to promise queue is only used for the update
+   count function. */
```
dlongley, line 11:
```suggestion
/* Note: Max size is 1. This promise queue is only used for the update
   count function. */
```

lib/cache/OperationWriter.js
```diff
@@ -7,13 +7,33 @@ const LRU = require('lru-cache');
...
+const lruCacheOptions = {max: 1, disposeOnSettle: true};
+const UPDATE_COUNT_PROMISE_QUEUE = new LruCache(lruCacheOptions);
```
dlongley, line 13: A size of 1 doesn't make sense because there may be multiple ledger nodes serviced by a single process -- and queue entries would be constantly evicted if just two were receiving ops.

lib/cache/OperationWriter.js
```diff
@@ -7,13 +7,33 @@ const LRU = require('lru-cache');
...
+   * database. In the extraordinary case, where the process dies between
+   * flushes, a maximum of 100 lost operations was deemed okay. A timeout of
+   * 1 second strikes the proper balance of ensuring timely flushes to the
+   * database, while also preventing overloading the database with constant
+   * requests.
+   *
+   * @param {object} options - Options to use.
+   * @param {object} options.ledgerNodeId - The `id` of the ledger node.
+   * @param {object} options.highWaterMark - The max number of records the
+   * writer should retain in memory before forcing a flush to the database.
```
dlongley, line 33:
```suggestion
   *   writer should retain in memory before forcing a flush to the database.
```

lib/cache/OperationWriter.js
```diff
@@ -7,13 +7,33 @@ const LRU = require('lru-cache');
...
+   * @param {object} options.timeout - The max amount of time in milliseconds
+   * records should be retained in memory before forcing a flush to the databse.
```
dlongley, line 35:
```suggestion
   *   records should be retained in memory before forcing a flush to the databse.
```

lib/cache/OperationWriter.js
```diff
@@ -7,13 +7,33 @@ const LRU = require('lru-cache');
...
 
 // Error Messages
 const DATABASE_QUEUE_LENGTH_ERROR =
   'Non-critical error: failed to cache length of operation queue on disk.';
 
 class OperationWriter {
+  /**
+   * Constructor for creating OperationWriters. The defaults for `timeout` and
+   * `highWaterMark` were chosen to ensure that at peak load, the operation
+   * writer will flush a 100 operations/second to the persisteted queue in the
```
dlongley, line 23:
```suggestion
   * writer will flush a 100 operations/second to the persisted queue in the
```

lib/cache/OperationWriter.js
```diff
@@ -7,13 +7,33 @@ const LRU = require('lru-cache');
...
+   * database. In the extraordinary case, where the process dies between
+   * flushes, a maximum of 100 lost operations was deemed okay. A timeout of
```
dlongley, line 25:
```suggestion
   * flushes, a maximum of 100 lost operations will occur. A timeout of
```

lib/cache/OperationWriter.js
```diff
@@ -34,8 +54,15 @@ class OperationWriter {
...
-    // tracks operation hashes for duplicate prevention
-    this.recentOperationHashes = new LRU({max: 1000, maxAge: 5 * 60 * 1000});
+    /* Tracks operations hashes for soft duplicate prevention within the same
+    process. This acts as a filter and first line of defense before inserting
+    operations into the queue in the database. If this soft check fails, then
+    the operation will try to be inserted into the database and yield a
+    DuplicateError. If a duplicate operation was inserted into the database and
+    did not already exist in the queue, then it will be filtered out when read
+    from the consensus worker.
+    */
```
dlongley, line 64:
```suggestion
    from the consensus worker. */
```

lib/cache/OperationWriter.js
```diff
@@ -101,14 +128,19 @@ class OperationWriter {
    * @param {object} options - Options to use.
    * @param {object} options.operation - The operation to write.
    * @param {object} options.meta - The metadata to write.
+   * @param {object} [options.forceFlush=false] - Force the queue to flush to
+   * the database after adding the operation.
```
dlongley, line 132:
```suggestion
   *   the database after adding the operation.
```

lib/cache/OperationWriter.js
```diff
@@ -34,8 +54,15 @@ class OperationWriter {
     // current stopped promise; resolves once stopped
     this.stopped = null;
     this._resolveStopped = null;
-    // tracks operation hashes for duplicate prevention
-    this.recentOperationHashes = new LRU({max: 1000, maxAge: 5 * 60 * 1000});
+    /* Tracks operations hashes for soft duplicate prevention within the same
```
dlongley, line 57:
```suggestion
    /* Tracks operation hashes for soft duplicate prevention within the same
```

lib/cache/OperationWriter.js
```diff
@@ -118,7 +150,11 @@ class OperationWriter {
           public: true,
         });
     }
-
+    /* Performs a soft duplicate check. This check does not prevent other
+    processes from accepting the operation or the case where the operation's
+    hash fell out of the in-memory LRU Cache. When the cheap and fast soft
+    check fails, there exists slower and robust checks when the operation
```
dlongley, line 156:
```suggestion
    check fails, there exist slower and robust checks when the operation
```

lib/worker/OperationQueue.js
```diff
@@ -81,19 +83,18 @@ module.exports = class OperationQueue {
...
-    // if the queue is not initialized or the head of the queue is at the end
-    // of the queue, then no next chunk of operations cached yet, so create one
-    // that will be cached when `getNextChunk` is called and its operations are
-    // retrieved
-    if(!(Array.isArray(this.queue) && (this.head + 1) < this.queue.length)) {
+    // if the queue is not initialized with the head of the queue before the end
+    // of the queue, then no next chunk of operations is cached yet, so create
+    // one that will be cached when `getNextChunk` is called and its operations
+    // are retrieved
+    const end = (this.queue || []).length - 1;
```
dlongley, line 90: We can avoid the unnecessary array creation:

```suggestion
    const end = this.queue ? this.queue.length - 1 : -1;
```

And did you mean for `end` to be `-1` or `0` in the case of no queue?

lib/worker/OperationQueue.js
```diff
@@ -81,19 +83,18 @@ module.exports = class OperationQueue {
...
+    if(!(Array.isArray(this.queue) && this.head < end)) {
```
dlongley, line 91: Can we simplify to an existence check on `this.queue`? It's never anything but `null`/`undefined` and an `Array`, right?

```suggestion
    if(!(this.queue && this.head < end)) {
```

lib/worker/OperationQueue.js
```diff
@@ -344,10 +345,10 @@ module.exports = class OperationQueue {
...
     const {ledgerNode} = this;
 
-    // Operations are validated at the latest block height of the ledger node
-    // before being accepted into the queue. We can return early when the ledger
-    // node has not advanced its block height, therefore preventing
-    // duplicating the execution of the same validation check.
+    /* Operations are validated at the latest block height of the ledger node
+	  before being accepted into the queue. We can return early when the ledger
+	  node has not advanced its block height, therefore preventing
+	  duplicating the execution of the same validation check. */
```
dlongley, line 351:
```suggestion
    /* Operations are validated at the latest block height of the ledger node
    before being accepted into the queue. We can return early when the ledger
    node has not advanced its block height, therefore preventing
    duplicating the execution of the same validation check. */
```

lib/worker/OperationQueue.js
dlongley: But the function is only called by the worker -- and the internal state of the queue should prevent another database call until what was retrieved last time is handled. I therefore don't understand the use case. Is this a remnant of the previous code that  did not have the non-currency assumption (the queue's API will only ever be called serially) built in?

lib/worker/OperationQueue.js
gannan08: Is the array not lazily created here? If the intent was to prevent allocating a new array entirely, your suggestion is a solution.

lib/worker/OperationQueue.js
gannan08: > ~~This is done to prevent multiple database calls when the function is called back-to-back in rapid succession~~. 

After reviewing the implementation of `_getHead` , it does have the non-concurrent call assumption. It returns `null` when a valid head could not be found. This means that it found 0 valid operations out of the set of operations in pulled from the database.

lib/cache/OperationWriter.js
```diff
@@ -0,0 +1,312 @@
...
+ */
+const database = require('bedrock-mongodb');
+const logger = require('../logger');
+const LRU = require('lru-cache');
+const {config, util: {BedrockError}} = require('bedrock');
+const {LruCache} = require('@digitalbazaar/lru-memoize');
+
+/* Note: Max size is 25. This promise queue is only used for the update
+   count function. We expect, in the worst case, no more than 25 ledger nodes
+   serviced by a single process updating their queue count. */
```
dlongley, line 12:
```suggestion
count function. We expect, in the worst case, no more than 25 ledger nodes
serviced by a single process updating their queue count. */
```

dlongley (approved):
Thanks!

lib/worker/OperationQueue.js
dlongley: Yes, that was my intent.

lib/operationWriter.js
```diff
@@ -0,0 +1,72 @@
...
+bedrock.events.on('bedrock.exit', async () => {
+  // flush all operation writers before exiting
+  OPERATION_WRITER_MAP.forEach(operationWriter => {
+    operationWriter.flush().catch(e => {
+      logger.error(`Error flushing operation writer.`, {
+        error: e,
+        ledgerNodeId: operationWriter.ledgerNodeId
+      });
+    });
+  });
```
dlongley, line 42:
```suggestion
  const promises = [];
  OPERATION_WRITER_MAP.forEach(operationWriter => {
    promises.push(operationWriter.flush().catch(e => {
      logger.error(`Error flushing operation writer.`, {
        error: e,
        ledgerNodeId: operationWriter.ledgerNodeId
      });
    }));
  });
  await Promise.all(promises);
```

dlongley:
Closing in favor of #262.

PR #259, opened 2021-06-29
Rename voters to peers and voterId to peerId.

aljones15:
Renames voters to peers in routes and removes `bedrock-rest` from this project.


test/aggregate.js
```diff
@@ -20,7 +20,7 @@ cursor = db['aa0d6eda-147d-469a-a4ef-075872701060-event'].aggregate([{
     restrictSearchWithMatch: {
```
dlongley, line 20: This whole file should be removed, it's obsolete.

lib/server.js
```diff
@@ -132,7 +132,7 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
 async function _handleNotification({req}) {
   const {peer} = req.body;
   const localPeerId = config.server.baseUri +
-    '/consensus/continuity2017/voters/' + req.params.voterId;
+    '/consensus/continuity2017/peers/' + req.params.peerId;
```
dlongley, line 135: If we add `encodeURIComponent` here and elsewhere, it will prepare us for using DIDs as peer IDs:

```suggestion
    `/consensus/continuity2017/peers/${encodeURIComponent(req.params.peerId)}`;
```

lib/server.js
```diff
@@ -88,7 +88,7 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
       // eslint-disable-next-line
       get: callbackify(async (req, res) => {
```
dlongley, line 89: We should remove this `linkedDataHandler` and make this is a regular route (this doesn't event accept linked data, so it's not correct). I suspect we don't need `brRest` at all in this package.

dlongley (changes requested):

test/aggregate.js
aljones15: addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/259/commits/18b0f6ad18ea4f9912621f59e27dda2d271a3c7c

lib/server.js
aljones15: Addressed here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/259/commits/b3e22465c61d6dfce42f13416ee92f0936f033fa

the rest of the calls on `brRest` are just `prefers.ld`. Should those stay or be removed along with `bedrock-rest` entirely?

lib/server.js
```diff
@@ -84,11 +83,10 @@ bedrock.events.on('bedrock-express.configure.routes', app => {
 
   app.post(
     routes.gossip, brRest.when.prefers.ld,
-    validate('continuity-server.gossip'), brRest.linkedDataHandler({
-      // eslint-disable-next-line
-      get: callbackify(async (req, res) => {
-        const localPeerId = config.server.baseUri +
-          '/consensus/continuity2017/voters/' + req.params.voterId;
+    validate('continuity-server.gossip'), asyncHandler(
+      (async (req, res) => {
```
dlongley, line 87: This needs some additional style fixes beyond this suggestion:

```suggestion
    validate('continuity-server.gossip'),
    asyncHandler(async (req, res) => {
```

lib/server.js
dlongley: Remove bedrock-rest entirely. I don't think the end points are even returning JSON-LD, it's JSON (that *may* have JSON-LD inside it) -- so it's not proper usage anyway.

aljones15: removed `bedrock-rest` here: https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/259/commits/c6139c0e4a9c9cb3ede38ff67e2be8c0121329dc

dlongley (approved):

PR #260, opened 2021-07-05
Upgrade bedrock ledger consensus continuity to witness pool.

msporny:


lib/witnesses.js
```diff
@@ -68,11 +68,11 @@ async function _getWitnesses({ledgerNode, blockHeight}) {
   // use witness selection method plugin to get witnesses for block
   const witnessSelectionMethod = _getWitnessSelectionMethod(
     {ledgerConfiguration});
-  if(witnessSelectionMethod.type !== 'electorSelection') {
+  if(witnessSelectionMethod.type !== 'witnessSelection') {
     throw new BedrockError(
       'Witness selection method is invalid.', 'InvalidStateError');
   }
-  const result = await witnessSelectionMethod.api.getBlockElectors({
+  const result = await witnessSelectionMethod.api.getBlockWitnesses({
```
dlongley, line 75: This:

```js
  const result = await witnessSelectionMethod.api.getBlockWitnesses({
    ledgerNode, ledgerConfiguration, latestBlockSummary, blockHeight
  });

  // only include `id` of witnesses, no extra data that may be set by
  // the selection algorithm
  const witnesses = result.electors.map(({id}) => id);
```

Should become this:

```js
  const result = await witnessSelectionMethod.api.getBlockWitnesses({
    ledgerNode, ledgerConfiguration, latestBlockSummary, blockHeight
  });

  const {witnesses} = result;
```

test/mocha/helpers.js
```diff
@@ -240,6 +240,27 @@ api.copyEvents = async ({from, to, useSnapshot = false}) => {
...
+      numWitnesses = 3 * f + 1;
+    }
+
+    // build the map of witnesses
+    const witnessPeers = peers.slice(0, numWitnesses);
+    for(const peer of witnessPeers) {
+      witnesses.set(peer.id, {id: peer.id});
+    }
+
+    return {witnesses};
```
dlongley, line 260:
```suggestion
    const witnesses = [];

    // calculate the number of witnesses
    let numWitnesses = 1;
    if(peers.values().length > 1) {
      const f = Math.floor((peers.values().length - 1) / 3);
      numWitnesses = 3 * f + 1;
    }

    // build the array of witnesses
    const witnesses = peers.slice(0, numWitnesses).map(({id}) => id);
    return {witnesses};
```

lib/witnesses.js
dlongley: And, really, the `Set` stuff should happen before the length check:

```js
  const witnesses = new Set(result.witnesses);
  // validate that `witnesses` is either size `1` or forms a `3f+1` size set
  if(witnesses.size !== 1 && witnesses.size % 3 !== 1) {
    throw new BedrockError(
      'Witnesses do not form a set of size "3f+1".', 'InvalidStateError');
  }
  return {witnesses};
```

test/mocha/090-multinode.js
```diff
@@ -86,8 +86,9 @@ describe('Multinode', () => {
     // override elector selection to force cycling and 3f+1
     before(() => {
       let candidates;
-      const witnessSelectionApi = brLedgerNode.use('MostRecentParticipants');
-      witnessSelectionApi.api.getBlockElectors = async ({blockHeight}) => {
+      const witnessSelectionApi = brLedgerNode.use('Continuity2017');
+      witnessSelectionApi.api._witnesses.getBlockWitnesses =
```
dlongley, line 90: Do not override the core function, override the plugin function. You'll need to replace `MostRecentParticipants` with whatever the new name is for the witness pool plugin.

```suggestion
      const witnessSelectionApi = brLedgerNode.use('MostRecentParticipants');
      witnessSelectionApi.api.getBlockWitnesses =
```

test/mocha/090-multinode.js
```diff
@@ -86,23 +86,26 @@ describe('Multinode', () => {
     // override elector selection to force cycling and 3f+1
     before(() => {
       let candidates;
-      const witnessSelectionApi = brLedgerNode.use('MostRecentParticipants');
-      witnessSelectionApi.api.getBlockElectors = async ({blockHeight}) => {
+      const witnessSelectionApi =
+        brLedgerNode.use('WitnessPoolWitnessSelection');
+      witnessSelectionApi.api.getBlockWitnesses =
+      async ({blockHeight}) => {
```
dlongley, line 92:
```suggestion
      witnessSelectionApi.api.getBlockWitnesses = async ({blockHeight}) => {
```

test/mocha/094-xblock-ws-wp.js
```diff
@@ -0,0 +1,230 @@
+/*!
```
dlongley, line 1: Let's name this `095-...` because there will be conflict with this already existing PR:

https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/257

This test may also be experiencing a hang / issues because it lacks the fix here:

https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/257/files#diff-9a0d33be264e030d26b5348338cf6d05e56023e29e1a81ef7ec7a9584b2bde45

But this may be avoidable once a witness pool doc is actually written.

msporny: Done in https://github.com/digitalbazaar/bedrock-ledger-consensus-continuity/pull/260/commits/26193890cb9ff732faad25d08151bdc25987ad74

dlongley (approved):

